PR #8, opened 2021-05-17
Modinreach 17

andrew-khodyka:
## Purpose
Create InnReachLocation JPA entity, liquibase migration, database relation


src/main/java/org/folio/innreach/config/JpaAuditingConfig.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Optional;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.domain.AuditorAware;
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing;
+
+@Configuration
+@EnableJpaAuditing(auditorAwareRef = "auditorAware")
```
dmtkachenko: I'd add modifyOnCreate=false to avoid setting modification date upon create

src/main/java/org/folio/innreach/config/JpaAuditingConfig.java
```diff
@@ -0,0 +1,18 @@
...
+public class JpaAuditingConfig {
+
+  @Bean
+  public AuditorAware<String> auditorAware() {
+    return () -> Optional.of("admin");
```
dmtkachenko, line 18: Is it always going to be admin as a current user?
As I understand it's a temp solution because current user has to be current auditor

Just for the reference, current users id is usually provided in `x-okapi-user-id` Http header

src/main/java/org/folio/innreach/domain/entity/base/Auditable.java
```diff
@@ -0,0 +1,40 @@
...
+@EntityListeners(AuditingEntityListener.class)
+public class Auditable {
+
+  @CreatedBy
+  @Column(name = "created_by")
+  private String createdBy;
+
+  @CreatedDate
+  @Column(name = "created_date")
+  private Date createdDate;
```
dmtkachenko: I'd suggest to use java.sql.Timestamp or even java.time.OffsetDateTime (to have timezone info) instead of java.util.Date

src/main/java/org/folio/innreach/domain/entity/InnReachLocation.java
```diff
@@ -0,0 +1,31 @@
...
+@EqualsAndHashCode(of = "code", callSuper = false)
+@ToString(callSuper = true)
+@Entity
+@Table(name = "inn_reach_location")
+public class InnReachLocation extends Auditable {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.AUTO)
+  private UUID id;
+  private String code;
```
dmtkachenko, line 29: From the requirements: 
> INN-Reach location code: limit five (5) lower-case alphanumeric characters
> INN-Reach location description: limited to 255 characters

How will the above limitations be supported?

src/test/resources/db/pre-populate-inn-reach-location-code.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO inn_reach_location (id, code, description, created_by, created_date, last_modified_by, last_modified_date)
```
dmtkachenko, line 1: Which test is this file used in?

src/main/resources/db/changelog/changes/v1.0.0/sql/2021-05-13-01__create-inn-reach-location_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE inn_reach_location
+(
+    id                 UUID         NOT NULL,
+    code               VARCHAR(5)   NOT NULL UNIQUE,
```
dmtkachenko: To enforce "lower-case alphanumeric characters" requirement you could use CHECK constraint as suggested here:
https://www.reddit.com/r/PostgreSQL/comments/2mal8j/is_there_an_alphanumeric_data_type/

src/main/java/org/folio/innreach/config/JpaAuditingConfig.java
andrew-khodyka: Yes, this is a temp solution. Will be changed when okapi integration is ready

src/main/java/org/folio/innreach/domain/entity/InnReachLocation.java
andrew-khodyka: We have the database constraint for the location code, also we'll add REST API validation for these values in the future

src/main/resources/db/changelog/changes/v1.0.0/sql/2021-05-13-01__create-inn-reach-location_table.sql
```diff
@@ -0,0 +1,10 @@
...
+    description        TEXT,
```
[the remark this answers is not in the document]
andrew-khodyka: Yes, my bad, I'll change it to VARCHAR(255)

src/test/resources/db/pre-populate-inn-reach-location-code.sql
andrew-khodyka: I used it just for the local test, but it will be used in REST API integration tests

src/main/java/org/folio/innreach/domain/entity/base/Auditable.java
andrew-khodyka: I was thinking about this, thanks for the adwise, will use java.time.OffsetDateTime

src/main/resources/db/changelog/changes/v1.0.0/sql/2021-05-13-01__create-inn-reach-location_table.sql
andrew-khodyka: I would prefer to add such kind of validation at the REST API level, since the mod-inn-reach service is the one owner of the database, so the REST API is the only entry point.

src/main/java/org/folio/innreach/config/JpaAuditingConfig.java
andrew-khodyka: Ok, will add this

src/main/java/org/folio/innreach/domain/entity/base/Auditable.java
```diff
@@ -20,21 +20,21 @@
 @ToString
 @MappedSuperclass
 @EntityListeners(AuditingEntityListener.class)
-public class Auditable {
+public class Auditable<T> {
```
andrew-khodyka, line 23: Made this class parameterized to be able to use a different type for user_id (maybe UUID would be better than String)

src/main/java/org/folio/innreach/config/JpaAuditingConfig.java
dmtkachenko: ok, thanks

src/main/java/org/folio/innreach/domain/entity/InnReachLocation.java
dmtkachenko: ok, agree

src/main/resources/db/changelog/changes/v1.0.0/sql/2021-05-13-01__create-inn-reach-location_table.sql
dmtkachenko: ok, let's do your way

src/main/java/org/folio/innreach/domain/entity/base/Auditable.java
dmtkachenko: 👍

dmtkachenko:
we need to add a DB test to verify repository and entity work as expected

dmtkachenko (approved):

PR #29, opened 2021-07-01
MODINREACH-84 Add proxy controller to get data directly from Inn Reach

andrew-khodyka:
## Purpose
Add proxy controller to get data directly from Inn Reach

## Approach

- Add controller to handle all requests with /d2r/** URI
- Add Inn Reach client to call Inn Reach API

## Pre-Merge Checklist:

  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
  - [x] Were any API paths or methods changed, added or removed?
  - [x] There are no breaking changes in this PR.
  

src/main/java/org/folio/innreach/external/client/feign/error/FeignErrorDecoder.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @Override
+  public Exception decode(String s, Response response) {
+    if (HttpStatus.valueOf(response.status()).equals(HttpStatus.UNAUTHORIZED)) {
+      log.debug("Can't get InnReach access token. CentralServer authentication failed with status: {}", response.status());
+      return new BadCredentialsException("Can't get InnReach access token. Key/Secret pair is not valid");
+    }
+
+    log.debug("Inn Reach call failed with status: {}", response.status());
+    return new InnReachException("Inn Reach call failed failed with status: {}");
```
dmtkachenko:
```suggestion
    return new InnReachException("Inn Reach call failed with status: " + response.status());
```

src/main/java/org/folio/innreach/external/service/impl/InnReachAuthServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+@Log4j2
+@Service
+public class InnReachAuthServiceImpl implements InnReachAuthService {
+
+  private final InnReachAuthClient innReachAuthClient;
+  private final Cache<String, AccessTokenDTO> accessTokenCache;
+
+  @Override
+  public AccessTokenDTO getAccessToken(AccessTokenRequestDTO tokenRequestDTO) {
+    var cachedAccessToken = accessTokenCache.getIfPresent(tokenRequestDTO.getCentralServerUri());
```
dmtkachenko, line 26: Would it be better to use Spring abstractions for caching like  @Cacheable with Caffeine as a provider?
Relevant documentation:
* https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.caching.provider
* https://docs.spring.io/spring-framework/docs/current/reference/html/integration.html#cache

Later on we can apply the same technic to other external calls like  querying mod-users

src/main/resources/application.yml
```diff
@@ -31,7 +31,7 @@ spring:
 folio:
   tenant:
     validation:
-      enabled: true
+      enabled: false
```
dmtkachenko: Could you explain why the validation is disabled? I don't think it's correct

src/main/resources/application.yml
```diff
@@ -8,7 +8,7 @@ spring:
     continue-on-error: true
     username: ${DB_USERNAME:postgres}
     password: ${DB_PASSWORD:postgres}
-    url: jdbc:postgresql://${DB_HOST:localhost}:${DB_PORT:5432}/${DB_DATABASE:db}
+    url: jdbc:postgresql://${DB_HOST:localhost}:${DB_PORT:5432}/${DB_DATABASE:postgres}
```
dmtkachenko: Was it a necessary change?

src/main/resources/application.yml
andrew-khodyka: Sorry, this is my fault. Changed for local testing, forgot to change back

src/main/resources/application.yml
andrew-khodyka: Sorry, this is my fault. Changed for local testing, forgot to change back

src/main/java/org/folio/innreach/external/service/impl/InnReachAuthServiceImpl.java
andrew-khodyka: Yes you are right. I was thinking about Spring Cache but decided to use Guava as a temporary solution and discuss Spring cache with you later.

src/main/java/org/folio/innreach/external/client/feign/error/FeignErrorDecoder.java
andrew-khodyka: Thanks, will fix this

dmtkachenko (approved):

dmtkachenko (commented):
New API URI has to be described in ModuleDescriptor-template.json otherwise it won't be accessible from UI
Not sure if it's possible to defined URI with a mask in the descriptor. Please consult RMB documentation regarding this point:
* https://github.com/folio-org/raml-module-builder/blob/master/README.md

If it's not possible then each endpoint has to described specifically like we did till this moment

PR #45, opened 2021-07-21
MODINREACH-12 Create a Data Model to Store a Mapping of Agencies to Folio Locations

oliinyko:
## Purpose
To facilitate circulation transactions, mod-inn-reach must create "temporary" instance, holding, and item records for items being borrowed from other agencies not hosted on the same local server. One of the required attributes of an item record is its permanent location. Based on the needs of the local server, it should be possible to configure what FOLIO location is assigned to items borrowed from these other agencies by mapping all agencies to a single FOLIO shelving location, each local server to a single location, each agency to its own dedicated location, or some combination of these. A default location must be mapped at the central server level, and will inherit to all local servers associated with that central server and all agencies associated with those local servers, unless otherwise explicitly configured. The list of local server and their associated agencies is supplied by the INN-Reach central server and should be referenced by their local server or agency codes (five-character, lower-case alpha-numeric strings). FOLIO locations should be referenced by Id.

## Approach
Create a Data Model to Store a Mapping of INN-Reach Agencies, Local Servers, and/or central servers to FOLIO Locations

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/domain/entity/AgencyLocationLscMapping.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private UUID locationId;
+
+  private UUID libraryId;
+
+  private String localServerCode;
+
+  @OneToMany(
+    cascade = CascadeType.ALL,
+    fetch = FetchType.EAGER,
```
andrew-khodyka: I believe the fetch type should be "LAZY", what do you think?

src/main/java/org/folio/innreach/domain/entity/AgencyLocationMapping.java
```diff
@@ -0,0 +1,44 @@
...
+  @Column(name = "central_server_id")
+  private UUID id;
+
+  private UUID locationId;
+
+  private UUID libraryId;
+
+  @OneToMany(
+    cascade = CascadeType.ALL,
+    fetch = FetchType.EAGER,
```
andrew-khodyka: The same as above

src/main/java/org/folio/innreach/domain/entity/AgencyLocationAcMapping.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.persistence.Table;
+import java.util.UUID;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@ToString(exclude = "localServerMapping")
+@Entity
+@Table(name = "agency_location_ac_mapping")
+public class AgencyLocationAcMapping implements Identifiable<UUID> {
```
andrew-khodyka: As far as I know, all entity classes must have audit columns, which means extending Auditable or AbstractEntity classes.

src/main/java/org/folio/innreach/domain/entity/AgencyLocationLscMapping.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import java.util.UUID;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@ToString(exclude = "centralServerMapping")
+@Entity
+@Table(name = "agency_location_lsc_mapping")
+public class AgencyLocationLscMapping implements Identifiable<UUID> {
```
andrew-khodyka: As far as I know, all entity classes must have audit columns, which means extending Auditable or AbstractEntity classes.

src/main/java/org/folio/innreach/domain/entity/AgencyLocationLscMapping.java
oliinyko: According to UI implementation, these entities are going to be fetched all together, I.e.
```
{
  "libraryId": "ef261062-e582-43d0-a1fc-c32dfca1da22",
  "locationId": "aa58c309-4522-4b46-8d1e-0396ee493460",
  "localServers": [
    {
      "localCode": "5publ",
      "localServerLibraryId": "ef261062-e582-43d0-a1fc-c32dfca1da22",
      "localServerLocationId": "aa58c309-4522-4b46-8d1e-0396ee493460",
      "agencyCodeMappings": [
        {
          "agencyCode": "5east",
          "libraryId": "70cf3473-77f2-4f5c-92c3-6489e65769e4",
          "locationId": "99b0d4e2-a5ec-46a1-a5ea-1080e609f969"
        }
      ]
    }
  ]
}
```
So there is no need to fetch them without child entities

src/main/java/org/folio/innreach/domain/entity/AgencyLocationAcMapping.java
oliinyko: The lcs/ac tables are child tables of aggency_location_mapping. The relation is defined with foreign key to parent mapping tables. They all are going to be updated simultaneously using single repository. That's why audit columns are only present in parent agency_location_mapping table. Let me know if you think these tables should be auditable anyway

src/main/java/org/folio/innreach/domain/entity/AgencyLocationLscMapping.java
andrew-khodyka: Ok, how about an update? If you need to update basic fields like localCode, libraryId or locationId OR add a new mapping to the agencyCodeMappings collection, you don't need to fetch the whole entity with the agencyCodeMappings, right? But you will :) since the fetch type is EAGER.

As far as I know, the best practice is to always use the LAZY fetch type and use JPQL JOIN FETCH queries when you need to get an entity with childs.

Please, correct me if I'm wrong.

src/main/java/org/folio/innreach/domain/entity/AgencyLocationAcMapping.java
andrew-khodyka: Not sure, but I guess all Entity classes should have audit columns. Let's ask @dmtkachenko what he think about this :)

src/main/java/org/folio/innreach/domain/entity/AgencyLocationLscMapping.java
oliinyko: Switched to use `Set` collections for nested entities instead of `List` as there are limitations joining multiple lists of entities in hibernate

src/main/java/org/folio/innreach/domain/entity/AgencyLocationAcMapping.java
dmtkachenko: I'd say it depends. 
If a child table is very simple entity, like collection of single values, and always included into parent entity, probably doesn't make sense to have separate metadata.
If a child table is more complex and can be concidered on its own then make sense to have a separate metadata. 

Another option is to think about how the entity (parent one and its children) used on UI. If everything always comes together then parent's metadata might be sufficient, and otherwise. If the option is unclear, just add metadata on both levels (parent + child)

oliinyko: Ok, thanks for the clarification. Added the audit columns to children tables.

src/main/java/org/folio/innreach/domain/entity/AgencyLocationMapping.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final String FETCH_BY_ID_POSTFIX = " WHERE am.id = :id";
+
+  public static final String FETCH_ONE_BY_ID_QUERY_NAME = "AgencyLocationMapping.fetchOne";
+
+  public static final String FETCH_ONE_BY_ID_QUERY = "SELECT DISTINCT am FROM AgencyLocationMapping AS am " +
+    "LEFT JOIN FETCH am.localServerMappings lsm " +
+    "LEFT JOIN FETCH lsm.agencyCodeMappings" + FETCH_BY_ID_POSTFIX;
+
+  @Id
+  @Column(name = "central_server_id")
```
dmtkachenko: please add `id` column to table as a synthetic PK

`central_server_id` should be represented as `CentralServer` field

dmtkachenko (approved):

PR #105, opened 2021-10-27
MODINREACH-92 D2IR Local Endpoint (Circulation): Create INN-Reach Patron Hold

oliinyko:
## Purpose
Create inventory virtual records for patron hold request
https://issues.folio.org/browse/MODINREACH-92

## Approach
- Add patron hold service
- Add move request logic
- Add reference data loader
- Upgrade spring boot version to 2.5.3 to enable mockity spying on repositories
- Add tests
- Add exception handler for d2ir controllers

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/controller/exception/D2irExceptionHandlerController.java
```diff
@@ -0,0 +1,26 @@
...
+import org.folio.innreach.dto.InnReachResponseDTO;
+
+@Log4j2
+@RestControllerAdvice(basePackages = "org.folio.innreach.controller.d2ir")
+public class D2irExceptionHandlerController {
+
+  @ExceptionHandler(Exception.class)
+  @ResponseStatus(HttpStatus.BAD_REQUEST)
+  public ResponseEntity<InnReachResponseDTO> handleException(Exception e) {
+    log.warn("Handling exception", e.getMessage());
```
dmtkachenko: What about printing out the stack trace? 
How would it be possible to quickly identify the place of error w/o having stack trace in the logs?

src/main/java/org/folio/innreach/domain/service/impl/CentralServerConfigurationServiceImpl.java
```diff
@@ -83,6 +88,16 @@ public CentralServerPatronTypesDTO getAllPatronTypes() {
         .totalRecords(csPatronTypes.size());
   }
 
+  @Override
+  public List<LocalServer> getLocalServers(UUID centralServerId) {
+    var agencies =
```
dmtkachenko: Replace the whole method body with: 

```suggestion
 return loadRecordPerServer(INN_REACH_LOCAL_SERVERS_URI, LocalServerAgenciesDTO.class,
              resp -> emptyIfNull(resp.getRight().getLocalServerList()), centralServerId);
```

src/main/java/org/folio/innreach/domain/service/impl/InventoryStorageServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+  @Override
+  public Instance queryInstanceByHrid(String instanceHrid) {
+    return getFirstItem(instanceStorageClient.queryInstanceByHrid(instanceHrid))
+      .orElseThrow(() -> new IllegalArgumentException("No instance found by hrid " + instanceHrid));
+  }
+
+  @Override
+  public ServicePointsClient.ServicePoint queryServicePointByCode(String locationCode) {
+    return getFirstItem(servicePointsClient.queryServicePointByCode(locationCode))
+      .orElseThrow(() -> new IllegalStateException("Service point is not found for pickup location code: " + locationCode));
```
dmtkachenko, line 41: In the previous method you use IllegalArgumentException to notify something is not found, here it is IllegalStateException -- doesn't look consistent

src/main/java/org/folio/innreach/domain/processor/PatronHoldInnReachCirculationProcessor.java
```diff
@@ -38,11 +45,17 @@ public InnReachResponseDTO process(String trackingId, String centralCode, Transa
...
+
+      patronHoldService.updateVirtualItems(innReachTransaction.get());
     } else {
       log.info("Transaction patron hold with trackingId [{}] and centralCode [{}] doesn't exist, create a new one...", trackingId, centralCode);
+
       InnReachTransaction newTransactionWithPatronHold = createTransactionWithPatronHold(trackingId, centralCode, transactionHold);
-      transactionRepository.save(newTransactionWithPatronHold);
+      var transaction = transactionRepository.save(newTransactionWithPatronHold);
+
+      patronHoldService.createVirtualItems(transaction);
```
dmtkachenko, line 58: According to the method definition, this operation is asynchronous, meaning patron hold transaction will be created and successful response returned to the client. 
What should happen with the transaction if async method fails at some step, e.g item is not created: 
- leave everything as is?
- delete the transaction?
- other option

src/main/java/org/folio/innreach/domain/service/impl/ReferenceDataLoader.java
```diff
@@ -0,0 +1,71 @@
...
+    new PathMatchingResourcePatternResolver(ReferenceDataLoader.class.getClassLoader());
+
+  private final InstanceContributorTypeClient instanceContributorTypeClient;
+  private final InstanceTypeClient instanceTypeClient;
+  private final JsonHelper jsonHelper;
+
+  public void loadRefData() {
+    try {
+      log.info("Loading reference data");
+      loadInstanceTypes();
```
dmtkachenko, line 47: not sure this is a good idea to populate data in other modules during tenant initialization

src/main/java/org/folio/innreach/repository/MaterialTypeMappingRepository.java
```diff
@@ -16,4 +16,6 @@
 
   Optional<MaterialTypeMapping> findOneByCentralServerIdAndMaterialTypeId(UUID centralServerId, UUID materialTypeId);
 
+  Optional<MaterialTypeMapping> findOneByCentralServerIdAndCentralItemType(UUID centralServerId, Integer centralItemType);
```
dmtkachenko: As I remember there might be several Folio material types mapped to the same central item type. Do you handle this situation properly?

oliinyko: Was using wrong service for this mapping. Switched to ItemTypeMappingService

src/main/java/org/folio/innreach/domain/processor/PatronHoldInnReachCirculationProcessor.java
oliinyko: The requirements are not clarified for error scenario. As soon as we have more details, we can update the logic. This shouldn't be a blocker I think

src/main/java/org/folio/innreach/domain/service/impl/ReferenceDataLoader.java
oliinyko: Updated service to run asynchronously with retries. This should make it much more reliable

src/main/java/org/folio/innreach/domain/processor/PatronHoldInnReachCirculationProcessor.java
dmtkachenko: well, the requirements indeed only describe the positive scenario, which doesn't mean we shouldn't think about negative cases. Otherwise we might face with serious problems when the code is on prod.
I suggest to at least ask PO about those potential negative scenarios and let him to decide whether it's a show stopper for the story or not

src/main/java/org/folio/innreach/domain/service/impl/ReferenceDataLoader.java
```diff
@@ -0,0 +1,100 @@
...
+
+  private <T> void load(String resourceDir, Class<T> resourceType,
+                        Function<T, ResultList<T>> loadExistingFunc,
+                        Consumer<T> createFunc) {
+    for (var res : getResources(resourceDir)) {
+      T rec = deserializeRecord(resourceType, res);
+
+      var existing = loadExistingFunc.apply(rec);
+
+      if (isEmpty(existing.getResult())) {
```
dmtkachenko, line 77: 👍

dmtkachenko (approved):

PR #109, opened 2021-11-03
MODINREACH-87

andrew-khodyka:
## Purpose
Handle Item Shipped Request for INN-Reach Patron Hold

## Approach

- Add circulation processor for PatronHold
- Add ITEM_SHIPPED CirculationOperation 
- Update inventory client

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/client/InventoryClient.java
```diff
@@ -22,4 +24,10 @@
 
   @GetMapping("/items?query=hrid=={hrId}")
   ResultList<InventoryItemDTO> getItemsByHrId(@PathVariable("hrId") String hrId);
+
+  @GetMapping("/items?query=barcode=={barcode}")
+  InventoryItemDTO getItemByBarcode(@PathVariable("barcode") String barcode);
```
oliinyko: The result of query is an array (most probably with single item). It should be wrapped to `ResultList`

src/main/java/org/folio/innreach/client/InventoryClient.java
```diff
@@ -22,4 +24,10 @@
...
+
+  @PutMapping("/items/{itemId}")
+  void updateItem(@PathVariable("itemId") UUID itemId, @RequestBody InventoryItemDTO inventoryItem);
```
oliinyko: I would suggest to use inventory-storage API instead. Inventory and inventory storage records have a bit different structure and I think it's better to keep using one API as long as it is reasonable. Right now we use inventory storage API for initial record contribution and for some of other logic and it looks much more suitable from backend perspective as it only loads identifiers of reference data without their details.

src/main/java/org/folio/innreach/domain/processor/ItemShippedInnReachCirculationProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  @Transactional
+  public InnReachResponseDTO process(String trackingId, String centralCode, TransactionHoldDTO transactionHold) {
+    var innReachTransaction = transactionRepository.findByTrackingIdAndCentralServerCode(trackingId, centralCode)
+        .orElseThrow(() -> new EntityNotFoundException(String.format(
+            "InnReach transaction with trackingId [%s] and centralCode [%s] not found", trackingId, centralCode)));
+
+    var itemBarcode = transactionHold.getShippedItemBarcode();
```
dmtkachenko: image
the request has "itemBarcode" property, from the other side TransactionHoldDTO contains field (and JsonProperty) named "shippedItemBarcode"..
How is the property from json request mapped to the class property?

src/main/java/org/folio/innreach/domain/processor/ItemShippedInnReachCirculationProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+  private final InventoryClient inventoryClient;
+
+  @Override
+  public boolean canProcess(String circulationOperationName) {
+    return CirculationOperation.ITEM_SHIPPED.getOperationName().equals(circulationOperationName);
+  }
+
+  @Override
+  @Transactional
+  public InnReachResponseDTO process(String trackingId, String centralCode, TransactionHoldDTO transactionHold) {
```
dmtkachenko: According to spec "itemshipped" request can contain only 2 fields:
- callNumber
- itemBarcode

TransactionHoldDTO, according to its schema, has several required fields:
`"required": [
    "transactionTime",
    "pickupLocation",
    "patronId",
    "patronAgencyCode",
    "itemAgencyCode",
    "itemId"
  ]`
Non of these fields are present in the request. How is it expected to make conversion from the request into the object if the mandatory fields are absent?

descriptors/ModuleDescriptor-template.json
```diff
@@ -326,6 +326,11 @@
             "manualblocks.collection.get"
           ]
         },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/inn-reach/d2ir/circ/{circulationOperationName}/{trackingId}/{centralCode}",
+          "permissionsRequired": ["inn-reach.d2ir.circulation.item.post"]
+        },
```
oliinyko, line 333: Item Shipped endpoint should use PUT method according to API documentation

src/main/java/org/folio/innreach/domain/processor/ItemShippedInnReachCirculationProcessor.java
andrew-khodyka: @dmtkachenko yes, my fault. I suppose we could use a single DTO to handle all types of request requests, in this case we need to remove all fields from the "required" block to make them "optional".

src/main/java/org/folio/innreach/client/InventoryClient.java
andrew-khodyka: @oliinyko thanks, I'll check

src/main/java/org/folio/innreach/domain/processor/ItemShippedInnReachCirculationProcessor.java
oliinyko: we could use generic dto `CirculationRequestDTO` without required fields and then map it to more specific DTO  (i.e. `TransactionHoldDTO`) where needed applying validation in runtime with `javax.validation.Validator`

oliinyko (approved):

src/main/java/org/folio/innreach/domain/processor/ItemShippedInnReachCirculationProcessor.java
andrew-khodyka: @dmtkachenko, I've changed the payloa, now it's CircualtionRequestDTO that contains the itemBarcode

PR #143, opened 2021-12-07
MODINREACH-153 Update MODINREACH for Compatibility with Optimistic Locking in Inventory

dmtkachenko:
## Purpose
Account for Optimistic locking enabled in Inventory during updates to Instance, Holdings and Item records, eg. as part of circulation flow for Patron Holds.
JIRA: [MODINREACH-153](https://issues.folio.org/browse/MODINREACH-153)

## Approach
- introduce update template to combine lookup operation (find) with modification operation (update)
- introduce basic service interface to describe essential operations (`create`, `update`, `find`) and special version of combined find + update (`changeAndUpdate`) which supports retries in case of "version conflict" exception. Retries are handled by Spring Retry library.
- split `InventoryService` into ItemService, HoldingsService, InstanceService, each extends the basic service interface
- modify calls from circulation service to "update" methods of Item/Holdings so that they can be recoverable from optimistic locking issues. if the issue cannot be automatically handled it's propagated outside as "failed" Inn-Reach response
- introduce feign decoder (`InventoryErrorDecoder`) to convert either Http 500 or Http 409 response from mod-inventory, which contains version conflict error details, into domain exception: `ResourceVersionConflictException`. This exception is configured as retryable in basic service interface
- apply the decoder to "Holdings" and "Inventory" feign clients

## Learning
Initially the idea was to override `update` method from `UpdateTemplate` in the descendant interfaces and apply retries configuration there. Thus there would be a single place to define retry policy. But this doesn't work and as a workaround `Retryable` has been placed twice to more specific `changeAndUpdate` methods

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/domain/service/BasicService.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Retryable(value = ResourceVersionConflictException.class,
+      maxAttemptsExpression = "#{${retryable-update.on-conflict.retry-attempts}}",
+      backoff = @Backoff(delayExpression = "#{${retryable-update.on-conflict.retry-interval-ms}}"),
+      listeners = {"retryMonitoringListener"})
+  default Optional<R> changeAndUpdate(K key, Supplier<? extends RuntimeException> notFoundExceptionSupplier,
+      UpdateOperation<R> change) {
+    return update(key, finder().toRequired(notFoundExceptionSupplier), change.andThen(this::update));
+  }
```
oliinyko, line 34: can't we simply relay on `Optional.orElseThrow` where needed instead of passing notFound handler as an argument?

dmtkachenko: that's what I've done initially, but Sonar detected a bug of not using a return value..
and on the second thought, I also decided that it's better to state during the method call that a result from finder cannot be empty, i.e. an exception has to be thrown

oliinyko: lets then pass it as the last param at least?

src/main/java/org/folio/innreach/domain/service/impl/ItemServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public InventoryItemDTO create(InventoryItemDTO item) {
+    return inventoryClient.createItem(item);
+  }
+
+  @Override
+  public InventoryItemDTO update(InventoryItemDTO item) {
+    inventoryClient.updateItem(item.getId(), item);
+    return item;
+  }
```
oliinyko, line 30: Is there any limitations why we can't receive and return the response of `inventoryClient.updateItem`?

src/main/java/org/folio/innreach/domain/service/BasicService.java
dmtkachenko: well, it's related to finder, which goes first. In this case the finder is not explicitly defined but it's assumed to be somewhere before the change. if there was a finder then it'd look something like
`changeAndUpdate(key, someFinder().toRequired(() -> throw 'not found'), changeOperation);`

Now omit the finder from the above statement and you will get the method signature I have defined

Your suggestion probably originates from `orElseThrow()` analogy , where it usually goes the last because it gets the result value. In case of `notFoundExceptionSupplier`, as I explained, it refers to look up method which is not the last one by the definition

src/main/java/org/folio/innreach/domain/service/impl/ItemServiceImpl.java
dmtkachenko: Yes there is. Whoever defined those methods should have checked the API documentation. According to it, PUT method returns no body but only Http 204 status
https://s3.amazonaws.com/foliodocs/api/mod-inventory/p/inventory.html#inventory_items__itemid__put

image

oliinyko: Yep, you are right)

oliinyko (approved):

PR #317, opened 2023-02-15
Modinreach 347 : Record Contribution Errors Should not Cause Contribution Jobs to Halt

arin-epm:
[MODINREACH-347](https://issues.folio.org/browse/MODINREACH-347)

## Purpose
Record Contribution Errors Should not Cause Contribution Jobs to Halt

## Approach
1. New runInitialContribution method created.
2. ConcurrentMessageListenerContainer implemented to consume Kafka messages through custom listener.
3. 503 response from D2IR server handled by creating custom ServiceSuspendedException.
4. Retry Kafka messages implemented for handling 503 response from D2IR server.

## Pre-Merge Checklist:

  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
  - [x] There are no breaking changes in this PR.
  

src/main/java/org/folio/innreach/batch/KafkaItemReader.java
```diff
@@ -57,11 +61,13 @@ public V read() {
     }
 
     if (consumerRecords.hasNext()) {
+      log.info("hasNext is true");
```
steveellis: This was for me debugging. You can remove.

src/main/java/org/folio/innreach/batch/KafkaItemReader.java
```diff
@@ -57,11 +61,13 @@ public V read() {
...
       ConsumerRecord<K, V> rec = consumerRecords.next();
       recordProcessor.accept(rec);
       partitionOffsets.put(new TopicPartition(rec.topic(), rec.partition()), rec.offset());
       return rec.value();
     } else {
+      log.info("hasNext is false");
```
steveellis: Same

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -5,23 +5,21 @@
...
 import static org.folio.innreach.batch.contribution.ContributionJobContextManager.endContributionJobContext;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.UUID;
-import java.util.concurrent.Future;
+import java.net.URI;
+import java.util.*;
```
steveellis, line 8: Let's always try to avoid wildcard imports.

src/main/java/org/folio/innreach/batch/contribution/IterationEventReaderFactory.java
```diff
@@ -69,7 +83,18 @@ private Deserializer<String> keyDeserializer() {
...
     return UUID.fromString(new String(rec.headers().lastHeader(ITERATION_JOB_ID_HEADER).value()));
+//    return UUID.fromString("fa34246b-86a6-4743-8d20-f368ef3242d7");
+  }
+
+  public KafkaUtil createKafkaConsumer() {
+
+    var consumerProperties = kafkaProperties.buildConsumerProperties();
+    consumerProperties.put(GROUP_ID_CONFIG, jobProperties.getReaderGroupId());
+
+    var topic = "folio.contrib.tester.innreach";
```
steveellis: Think this topic is just for testing. The prod topic is different.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,48 +68,66 @@ public class ContributionJobRunner {
 
   private static final List<UUID> runningInitialContributions = Collections.synchronizedList(new ArrayList<>());
 
-  @Async
-  public Future<Void> runInitialContributionAsync(UUID centralServerId, String tenantId, UUID contributionId, UUID iterationJobId) {
+  public void runInitialContributionAsync(UUID centralServerId, String tenantId, UUID contributionId, UUID iterationJobId) {
```
steveellis: Should rename this method since it's no longer async.

src/main/java/org/folio/innreach/external/client/feign/error/InnReachFeignErrorDecoder.java
```diff
@@ -20,6 +21,11 @@ public Exception decode(String methodKey, Response response) {
       return new BadCredentialsException("Can't get InnReach access token. Key/Secret pair is not valid");
     }
 
+    if (HttpStatus.valueOf(response.status()).equals(HttpStatus.SERVICE_UNAVAILABLE)) {
+      log.debug("CentralServer is suspended for contribution");
+      return new ServiceSuspendedException("CentralServer is suspended for contribution. Try again later");
+    }
+
```
steveellis, line 28: This is good, but I think this is only going to handle the exception if we are using version 3. We also need to check for the string that Justin provided in the response body for version 2.

src/main/resources/application.yml
```diff
@@ -5,11 +5,11 @@ spring:
     name: mod-inn-reach
   datasource:
     username: ${DB_USERNAME:postgres}
-    password: ${DB_PASSWORD:postgres}
+    password: ${DB_PASSWORD:root}
```
steveellis: This seems unrelated to this PR. Is it?

src/main/resources/application.yml
```diff
@@ -5,11 +5,11 @@ spring:
...
     url: jdbc:postgresql://${DB_HOST:localhost}:${DB_PORT:5432}/${DB_DATABASE:db}
   jpa:
     hibernate:
-      ddl-auto: none
+      ddl-auto: update
```
steveellis: Same as above. Is this related to this PR?

src/main/resources/application.yml
```diff
@@ -168,3 +168,6 @@ kafka:
       topic-pattern: (${environment}\.)(.*\.)inventory\.holdings-record
       concurrency: ${KAFKA_EVENTS_CONCURRENCY:2}
       data-type: org.folio.innreach.dto.Holding
+  backoff:
+    interval: 30000
+    max_failure: 5
```
steveellis: Thinking ahead to how we're going to want to do the same thing for the instances of consumers for _ongoing contributions_. If we set these globally in the app yaml are they going to apply to _every instance_ of our use of the  `ConcurrentMessageListenerContainer`?

Is the retry and max attempts different for initial contrib vs ongoing is what I'm asking. At first I thought yes they are, but now I'm not so sure. Perhaps they should be exactly alike.
 
But these values will not suffice for production I think. We need for it to retry for as long as 12 hours (I'm thinking), and as we have been saying the interval **must** be > than the default timeout from okhttp3 unless I'm missing something. To get the max_failure figure out how many times of the interval will get us to 12 hours.

src/main/java/org/folio/innreach/batch/contribution/IterationEventReaderFactory.java
```diff
@@ -41,13 +44,24 @@ public class IterationEventReaderFactory {
...
+  @Qualifier("retryInterval")
+  private final Long interval;
+
+  @Qualifier("retryMaxAttempts")
+  private final Long maxAttempts;
+
+  @Qualifier("itemExceptionListener")
+  private final ContributionExceptionListener contributionExceptionListener;
+
   public KafkaItemReader<String, InstanceIterationEvent> createReader(String tenantId) {
```
steveellis, line 54: Let's remove this method entirely and its tests. Why keep it around?

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,48 +68,66 @@ public class ContributionJobRunner {
...
-    runInitialContribution(context);
+    KafkaUtil tempKafkaConsumer = itemReaderFactory.createKafkaConsumer();
+
+    CustomAckMessageListener customAckMessageListener = new CustomAckMessageListener(new ContributionProcessor(this), context, new Statistics());
 
-    return new AsyncResult<>(null);
+    tempKafkaConsumer.startOrCreateConsumer(customAckMessageListener);
   }
 
+  @Deprecated
```
steveellis: Why deprecate? Just remove and if it is used in tests, remove those too.

src/main/java/org/folio/innreach/util/KafkaUtil.java
```diff
@@ -0,0 +1,114 @@
...
+import java.net.SocketTimeoutException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG;
+
+@Log4j2
+@Setter
+@RequiredArgsConstructor
+public class KafkaUtil {
```
steveellis: I know this name is taken from the sample code, but it doesn't really work well for our purposes. How about something like `InitialContributionJobConsumerContainer`? It's got stuff in it that is specific to the init contribution job like the error handlers so it probably won't be reused any other place.

So remove from util directory obviously and put in a better place.

src/main/java/org/folio/innreach/util/KafkaUtil.java
```diff
@@ -0,0 +1,114 @@
...
+  private final Map<String,Object> consumerProperties;
+  private final String topic;
+  private final Deserializer<String> keyDeserializer;
+  private final Deserializer<InstanceIterationEvent> valueDeserializer;
+
+  private final Long interval;
+
+  private final Long maxAttempts;
+
+  private final int THREAD_SIZE = 2;
```
steveellis: I think it's fine to hardcode this for our first test run with prod data. 2 should give us roughly twice as good perf and likely won't overtax the container.

src/main/java/org/folio/innreach/util/KafkaUtil.java
```diff
@@ -0,0 +1,114 @@
...
+      contributionExceptionListener.logWriteError(exception, record.value().getInstanceId());
+      log.info("Stopping consumer topic: {} after retry exhaustion", consumerRecord.topic());
+      stopConsumer(consumerRecord.topic());
+    }, fixedBackOff);
+    errorHandler.addRetryableExceptions(ServiceSuspendedException.class);
+    errorHandler.addRetryableExceptions(SocketTimeoutException.class);
+    return errorHandler;
+  }
+
+  public void startOrCreateConsumer(Object messageListner) {
```
steveellis: For methods that attempt to do something (here start if it isn't already started) I like to prepend a `try` on the method name. So here I think this method should simply be `tryStart`. I think the word `consumer` is confusing/misleading here. We're basically starting the container which starts one or more consumers. If we rename the class as I recommend above, I think `tryStart` is clear.

src/main/java/org/folio/innreach/batch/contribution/IterationEventReaderFactory.java
```diff
@@ -69,7 +83,18 @@ private Deserializer<String> keyDeserializer() {
   }
 
   private static UUID getJobId(ConsumerRecord<String, InstanceIterationEvent> rec) {
+    // TODO Comment this back in after testing
     return UUID.fromString(new String(rec.headers().lastHeader(ITERATION_JOB_ID_HEADER).value()));
+//    return UUID.fromString("fa34246b-86a6-4743-8d20-f368ef3242d7");
+  }
+
+  public KafkaUtil createKafkaConsumer() {
```
steveellis: Let's rename this method to `createInitialContributionConsumerContainer` since it can be more than one and I'm recommending a rename of `KafkaUtil` to be more in line with our purposes.

src/main/java/org/folio/innreach/batch/contribution/IterationEventReaderFactory.java
steveellis: See how the old `createReader` constructs the topic:
```
var topic = String.format("%s.%s.%s", folioEnv.getEnvironment(), tenantId, jobProperties.getReaderTopic());
```

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
steveellis: How about `startInitialContribution` since `runInitialContribution` is already taken.

src/main/java/org/folio/innreach/batch/contribution/service/CustomAckMessageListener.java
```diff
@@ -0,0 +1,28 @@
...
+
+import lombok.AllArgsConstructor;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.folio.innreach.batch.contribution.ContributionJobContext;
+import org.folio.innreach.domain.dto.folio.inventorystorage.InstanceIterationEvent;
+import org.springframework.kafka.listener.AcknowledgingMessageListener;
+import org.springframework.kafka.support.Acknowledgment;
+
+@AllArgsConstructor
+public class CustomAckMessageListener implements AcknowledgingMessageListener<String, InstanceIterationEvent> {
```
steveellis: I think this is no longer an "acknowledging message listener" because it no longer acks (commits). I think two things should happen here:
* Go back to the sample code. It had an implementation that wasn't acknowledging and implement that instead here. Of course testing that it works.
* Rename this class to something more meaningful for our purposes like `InitialContributionMessageListener`.

src/main/java/org/folio/innreach/external/client/feign/error/InnReachFeignErrorDecoder.java
arin-epm: That part is handled in InnReachResponse object where we are getting 200 response while contributing & decontributing

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,114 @@
...
+  private final Map<String,Object> consumerProperties;
+  private final String topic;
+  private final Deserializer<String> keyDeserializer;
+  private final Deserializer<InstanceIterationEvent> valueDeserializer;
+
+  private final Long interval;
+
+  private final Long maxAttempts;
+
+  private final int THREAD_SIZE = 2;
```
heorhiiii: _in general code could be prettier to increase its readability and maintainability_. E.g.: even within current example there are two omissions: 1) for such format it should be static final field; 2) for increasing meaning - it could be something like: "concurrent consumers amount" or "concurrent container size" as once noticed THREAD_SIZE constant devs who will maintain the codebase will likely imply Thread Pool, but it's just a property for concurrency of ConcurrentMessageListenerContainer.

src/main/java/org/folio/innreach/batch/KafkaItemReader.java
```diff
@@ -46,6 +49,7 @@
   public void open() {
     if (kafkaConsumer == null) {
       kafkaConsumer = new KafkaConsumer<>(consumerProperties, keyDeserializer, valueDeserializer);
+     // new DefaultKafkaConsumerFactory<>(kafkaConsumer)
```
steveellis: Should this line be removed?

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,114 @@
...
+  private final Map<String,Object> consumerProperties;
+  private final String topic;
+  private final Deserializer<String> keyDeserializer;
+  private final Deserializer<InstanceIterationEvent> valueDeserializer;
+
+  private final Long interval;
+
+  private final Long maxAttempts;
+
+  private static final int CONCURRENT_CONTAINER_SIZE = 2;
```
steveellis: Maybe just `CONCURENCY` would be better. Not sure if "size" really makes much sense here.

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,114 @@
...
+
+    log.info("container is null");
+
+    ContainerProperties containerProps = new ContainerProperties(topic);
+
+    containerProps.setPollTimeout(100);
+    Boolean enableAutoCommit = (Boolean) consumerProperties.get(ENABLE_AUTO_COMMIT_CONFIG);
+//    if (!enableAutoCommit) {
+//      containerProps.setAckMode(ContainerProperties.AckMode.MANUAL_IMMEDIATE);
+//    }
```
steveellis: Should probably remove commented out code.

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,114 @@
...
+    container = new ConcurrentMessageListenerContainer<>(factory, containerProps);
+
+    container.setupMessageListener(messageListner);
+    container.setCommonErrorHandler(errorHandler());
+
+//    if (concurrency == 0) {
+//      container.setConcurrency(1);
+//    } else {
+//      container.setConcurrency(concurrency);
+//    }
```
steveellis: No longer needed so should remove.

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,114 @@
...
+    consumersMap.put(topic, container);
+
+    log.info("created and started kafka consumer for topic {}", topic);
+  }
+
+  public static void stopConsumer(final String topic) {
+    log.info("stopping consumer for topic {}", topic);
+    ConcurrentMessageListenerContainer<String, InstanceIterationEvent> container = consumersMap.get(topic);
+    container.stop();
+    log.info("consumer stopped!!");
```
steveellis: Probably should add topic name to log message like above.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
       .centralServerId(centralServerId)
       .tenantId(tenantId)
       .build();
 
-    runInitialContribution(context);
+    totalRecords = numberOfRecords;
+
+    recordsProcessed.put(tenantId, 0);
+
+    InitialContributionJobConsumerContainer tempKafkaConsumer = itemReaderFactory.createInitialContributionConsumerContainer(tenantId);
```
steveellis: This variable name should probably be `container` or similar.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
-          }
-        }
-      });
+
+    var iterationJobId = context.getIterationJobId();
+
+    if (event == null) {
+      log.info("Exiting kafka message reader for contribution {} and job {}", contributionId, iterationJobId);
+      return;
+    }
```
steveellis, line 128: I don't think we want this block any more do we? Or maybe we want to keep the null check (so it doesn't blow up if it gets one), but change the log message to something more accurate to what would happen in this case.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
+  public void simulateContribution() {
+    makeSimulatedRequest();
+//    s.addRecordsTotal(1);
+  }
+
+  private void simulateContributionItems() {
+    // For now just make this one.
+    makeSimulatedRequest();
+//    s.addRecordsTotal(1);
+  }
```
steveellis: I think these two methods can be removed just to clean things up.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
+      String postBody = "somerandomstring"; // Should probably make this a random length.
+      URI testServer = URI.create("http://localhost:8080");
+      var res = testClient.makeTestRequest(testServer, postBody);
+      log.info("Response from test server: {}", res);
+    } catch (Exception e) {
+      log.warn("Error while simulating request: {} {}", e.getMessage(), e.getStackTrace());
+    } finally {
+      s.addRecordsProcessed(1);
+    }
+  }
```
steveellis: I think this method can be removed. It's not useful anymore is it? We have the one below which we can keep.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
+      }
+
+    } catch (ServiceSuspendedException ex) {
+      throw new ServiceSuspendedException(ex.getMessage());
+    } catch (Exception e) {
+      log.warn("Error while simulating request: {} {}", e.getMessage(), e.getStackTrace());
+    }
+  }
+
+  public Instance simulateLoadingInstance(Statistics s) {
```
steveellis: Can this method be removed?

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -64,57 +73,125 @@ public class ContributionJobRunner {
...
+      var res = testClient.makeTestRequest(testServer, postBody);
+      log.info("Response from test server: {}", res);
+    } catch (Exception e) {
+      log.warn("Error while simulating request: {} {}", e.getMessage(), e.getStackTrace());
+    } finally {
+      s.addRecordsProcessed(1);
+    }
+  }
+
+  private void makeSimulatedRequest() {
```
steveellis: Let's keep this method around since it makes testing a lot easier.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionProcessor.java
```diff
@@ -0,0 +1,22 @@
...
+public class ContributionProcessor implements IMessageProcessor{
+
+  private final ContributionJobRunner contributionJobRunner;
+
+  @Override
+  public void processMessage(String key, InstanceIterationEvent event, ContributionJobContext context, ContributionJobContext.Statistics statistics, String topic) {
+    System.out.println("Message is:"+ event.toString());
+    log.info("Processing initial contribution job {}", context);
+    contributionJobRunner.runInitialContribution(context, event, statistics, topic);
+//    contributionJobRunner.simulateContribution();
```
steveellis: Maybe add a comment just above this commented out line that this is for testing using a local kafka and a local simulated d2ir server.

src/main/java/org/folio/innreach/batch/contribution/service/CustomAckMessageListener.java
steveellis: Don't think this is resolved Arin. Should this be an `AcknowledgingMessageListener` or just a `MessageListener` implementation. [This](https://github.com/bkatwal/kafka-util/blob/master/src/main/java/com/bkatwal/kafka/impl/CustomMessageListener.java#L10) is the sample code I'm referring to.

On line 26 you're no longer acknowledging. Because we have retry, we don't need to commit/ack, right? It's auto committing now, and retry is smart enough to do commit only when the retry succeeds. Have we verified that it properly retries and only commits on success?

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionMessageListener.java
```diff
@@ -0,0 +1,28 @@
...
+
+import lombok.AllArgsConstructor;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.folio.innreach.batch.contribution.ContributionJobContext;
+import org.folio.innreach.domain.dto.folio.inventorystorage.InstanceIterationEvent;
+import org.springframework.kafka.listener.AcknowledgingMessageListener;
+import org.springframework.kafka.support.Acknowledgment;
+
+@AllArgsConstructor
+public class InitialContributionMessageListener implements AcknowledgingMessageListener<String, InstanceIterationEvent> {
```
steveellis: See my comment on previous versions. Should this be implementing this interface and not the `MessageListener` one?

src/main/java/org/folio/innreach/domain/service/impl/RecordContributionServiceImpl.java
```diff
@@ -121,9 +132,19 @@ private InnReachResponse verifyBibContribution(UUID centralServerId, String bibI
...
-    Assert.isTrue(response.isOk(), "Unexpected items contribution response: " + response);
-    return response;
+    try {
+      var response = irContributionService.contributeBibItems(centralServerId, bibId, BibItemsInfo.of(bibItems));
+      if (!response.getErrors().isEmpty()) {
+        var error = response.getErrors().get(0).getReason();
+        if (error.contains("Contribution to d2irm is currently suspended")) {
+          throw new ServiceSuspendedException("Contribution to d2irm is currently suspended");
+        }
+      }
```
steveellis: This block is duplicated with the block above in `contributeBib`. Code duplication is always bad. Abstract this block into a method on the response object called `checkServiceSuspension` which throws our exception.

src/main/java/org/folio/innreach/domain/service/impl/RecordContributionServiceImpl.java
```diff
@@ -109,9 +110,19 @@ private void logItemTransformationError(Item item, Exception e) {
...
+      Assert.isTrue(response.isOk(), "Unexpected contribution response: " + response);
+      return response;
+    } catch (ServiceSuspendedException ex) {
+      throw new ServiceSuspendedException(ex.getMessage());
```
steveellis: So we throw an exception, catch the same exception, and re-throw it. What would be the purpose of doing that here? Throwing it once is all you need. There's no need for `try catch` here or in the other methods that do this.

src/main/java/org/folio/innreach/external/client/feign/error/InnReachFeignErrorDecoder.java
steveellis: But why are handling this in two different places? I understand that one is a response code and one is a response body, but they are both part of the same response. Shouldn't error handling be consolidated in one place? See my comments below as well about code duplication.

src/main/java/org/folio/innreach/external/service/impl/InnReachContributionServiceImpl.java
```diff
@@ -46,32 +47,54 @@ public InnReachResponse contributeBib(UUID centralServerId, String bibId, BibInf
...
+      if (!response.getErrors().isEmpty()) {
+        var error = response.getErrors().get(0).getReason();
+        if (error.contains("Contribution to d2irm is currently suspended")) {
+          throw new ServiceSuspendedException("Contribution to d2irm is currently suspended");
+        }
+      }
+      return response;
+    } catch (ServiceSuspendedException ex) {
+      throw new ServiceSuspendedException(ex.getMessage());
+    }
```
steveellis, line 99: This is nearly the exact same code block as in `deContributeBib` above. Consolidate into a shared method.

src/main/resources/application.yml
steveellis: How has this been resolved? I see that you have increased the interval and the max failure, but how do the new values you have chosen address my concerns?

src/main/resources/application.yml
```diff
@@ -168,3 +168,6 @@ kafka:
       topic-pattern: (${environment}\.)(.*\.)inventory\.holdings-record
       concurrency: ${KAFKA_EVENTS_CONCURRENCY:2}
       data-type: org.folio.innreach.dto.Holding
+  backoff:
+    interval: 3600000
+    max_failure: 12
```
steveellis: These values are almost certainly not correct. Without these values being exactly right the entire solution that we have come up with won't work. Please read the documentation carefully for these values and come up with something that will work for our purposes.

As I have stated elsewhere: we want it to retry for at least 12 hours and retry needs to not conflict with okhttp3.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionMessageListener.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Override
+  public void onMessage(
+    ConsumerRecord<String, InstanceIterationEvent> consumerRecord) {
+
+    // process message
+    iMessageProcessor.processMessage(consumerRecord.key(), consumerRecord.value(), context, statistics, consumerRecord.topic());
+
+    // commit offset
+//    acknowledgment.acknowledge();
```
steveellis: Think this commented out bit can be removed now.

src/main/java/org/folio/innreach/batch/contribution/IterationEventReaderFactory.java
```diff
@@ -72,4 +80,14 @@ private static UUID getJobId(ConsumerRecord<String, InstanceIterationEvent> rec)
...
   }
 
+  public InitialContributionJobConsumerContainer createInitialContributionConsumerContainer(String tenantId) {
+
+    var consumerProperties = kafkaProperties.buildConsumerProperties();
+    consumerProperties.put(GROUP_ID_CONFIG, jobProperties.getReaderGroupId());
+
+    var topic = String.format("%s.%s.%s",
+      folioEnv.getEnvironment(), tenantId, jobProperties.getReaderTopic());
+    return new InitialContributionJobConsumerContainer(consumerProperties,topic,keyDeserializer(),valueDeserializer(), retryConfig.getInterval(), retryConfig.getMaxAttempts(), contributionExceptionListener);
```
steveellis: Need a space between the method parameters

src/main/java/org/folio/innreach/batch/contribution/service/ContributionProcessor.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Log4j2
+@AllArgsConstructor
+public class ContributionProcessor implements IMessageProcessor{
+
+  private final ContributionJobRunner contributionJobRunner;
+
+  @Override
+  public void processMessage(InstanceIterationEvent event, ContributionJobContext context, ContributionJobContext.Statistics statistics, String topic) {
+    System.out.println("Message is:"+ event.toString());
```
steveellis: Don't forget to remove this

Vignesh-kalyanasundaram:
this PR is not required. So, closing this PR.

PR #323, opened 2023-03-09
FOLSPRINGB-95 - non-public beginFolioExecutionContext avoids wrong tenant/user

TarasSpashchenko:
## Purpose
To eliminate cases when a thread runs with a wrong tenant and wrong user.

Manually calling beginFolioExecutionContext and endFolioExecutionContext is error-prone because one or both of them can be forgotten. If forgotten the unit tests doesn't catch this.

## Approach
Two options are available that enforce that beginFolioExecutionContext is called at the beginning of some task and endFolioExecutionContext is always called afterward:

https://github.com/folio-org/folio-spring-base/blob/v6.0.1/src/main/java/org/folio/spring/scope/FolioExecutionScopeExecutionContextManager.java#L71-L93

```
import static org.folio.spring.scope.FolioExecutionScopeExecutionContextManager.getRunnableWithCurrentFolioContext;

executor.execute(getRunnableWithCurrentFolioContext(() -> /* some stuff */));

```
https://github.com/folio-org/folio-spring-base/blob/v6.0.1/src/main/java/org/folio/spring/scope/FolioExecutionContextSetter.java

```
try (var x = new FolioExecutionContextSetter(currentFolioExecutionContext)) {
  // some stuff
}
```


src/main/java/org/folio/innreach/domain/service/impl/SystemUserAuthService.java
```diff
@@ -58,8 +54,7 @@ public void setupSystemUser() {
   }
 
   public String loginSystemUser(SystemUser systemUser) {
-    return executeWithinContext(contextBuilder.forSystemUser(systemUser), () -> {
-
+    try (var context = new FolioExecutionContextSetter(contextBuilder.forSystemUser(systemUser))) {
```
steveellis, line 57: It looks like sonar is complaining about this new try-with-resources block. Looks like it is considering the whole `try` block new code and it looks like the method has code coverage, but perhaps not the optional bit or the thrown exception. It would be nice if sonar was more specific here.

steveellis (approved):
Don't think my comment needs to be resolved Taras. Probably coverage could be improved but the old code had the same problem.

PR #324, opened 2023-03-14
MODINREACH-347: Record Contribution Errors Should not Cause Contribution Jobs to Halt

Arghya22:
MODINREACH-347 (https://issues.folio.org/browse/MODINREACH-347)

## Purpose
1. When errors occur during record contribution, the error should be logged, the record skipped, and the overall contribution job should continue running until all instances have been processed for contribution. In the current implementation, the job is completing after a few thousand records have been processed after receiving an unrecognized error code from the central.

## Approach
1. Redesign of existing Kafka Consumer 
2. Implemented concurrency.
3. Implemented Kafka retry along with Spring retry. 
4. Implemented policy for when what retry should trigger. 
5. New counting logic added which is used to stop Kafka consumer from processing in case of total completion of records or retry exhaustions.
6. Restructure of existing business logic in order to fit the current implementation.
7. Test cases added and modified accordingly.


steveellis:
@Arghya22 rather than comment every line, I would prefer if you could go through it again line by line and please make the following changes:
* Remove necessary characters from log messages (for example: `::`, `---`, `->`, `-->>`, `=`  and probably others). This just clutters things up and doesn't add value.
* If a log message has a parameter, introduce that consistently with a colon and a space `log.info(Some message: {}, param)`. You do this some places but not others.
* Always use  the `{}` syntax rather than concatenating the string with `+` when you have a parameter or parameters in your log messages.
* Check your use of spaces throughout. For example, method parameters should always have a comma and a space between them. Sometimes you have no spaces (or some with spaces and some without). Sometimes you introduce conditions without a space like `if(something`. 
* Remove unnecessary empty lines.
* Some of the `info` messages that you have seem more like `debug` messages. Either change them to `debug` if you think they might be useful in the future, or remove them entirely.
* Look at each of your log messages and imagine yourself trying to debug a production system. Make sure each message succinctly and describes with proper grammar what is being logged and if there is something like a param or params that can uniquely identify it, add that. For example you have two `InnReachConnectionException occur---`. These are each separate things.

steveellis:
Also @Arghya22 please go through the [code smells report on sonar](https://sonarcloud.io/project/issues?resolved=false&pullRequest=317&id=org.folio%3Amod-inn-reach) and fix as many as you can. These are almost always worth addressing. As for the ones about code not being addressed in unit tests, let's discuss that. It may be that we can ignore those for now since you're at 80% coverage, but the rest are easy and necessary fixes.

src/main/java/org/folio/innreach/config/props/InnReachRetryPolicy.java
```diff
@@ -0,0 +1,45 @@
...
+
+      this.setExceptionClassifier((Classifier<Throwable, RetryPolicy>) classifiable -> {
+
+        if (classifiable instanceof ServiceSuspendedException) {
+            log.info("classifiable is of ServiceSuspendedException");
+          return new NeverRetryPolicy();
+        }
+        else if (classifiable instanceof FeignException) {
+          log.info("classifiable is of FeignException");
+          return new NeverRetryPolicy();
```
steveellis, line 31: Doesn't putting `FeignException` here mean that all non 200 exceptions from the HTTP client will not retry on the Spring level and always bubble up to the Kafka level? I think this will still _work_ but my understanding was that you wanted for verification (which is a 404 presumably when it fails to verify) to happen on the Spring level, not the Kafka level. Is that still the case?

If you look at [the nested classes for this exception](https://javadoc.io/doc/io.github.openfeign/feign-core/11.0/feign/FeignException.html) you'll notice that it's pretty much any non-200 HTTP response that will be an instance of `FeignException`.

src/main/java/org/folio/innreach/domain/service/impl/RecordContributionServiceImpl.java
```diff
@@ -109,22 +116,47 @@ private void logItemTransformationError(Item item, Exception e) {
...
+      if(errorResponse!=null && errorResponse.getMessages()!=null && !errorResponse.getMessages().isEmpty()) {
+        errorMessages = errorResponse.getMessages().get(0);
+      }
+      log.info("checkServiceSuspension:: error is : {}",error);
+      if (error.contains(CONTRIBUTION_TO_D2IRM_IS_CURRENTLY_SUSPENDED)) {
+        log.info("ServiceSuspendedException occur");
+        throw new ServiceSuspendedException(CONTRIBUTION_TO_D2IRM_IS_CURRENTLY_SUSPENDED);
+      }
+      if(errorMessages.contains(CONNECTIONS_ALLOWED_FROM_THIS_SERVER)) {
+        log.info("InnReachConnectionException occur");
```
steveellis: This info log message here and the one above should not be the same. Otherwise we won't be able to tell in logging which one happened.

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Map;
+
+import static org.folio.innreach.batch.contribution.ContributionJobContextManager.getContributionJobContext;
+
+@Log4j2
+@Setter
+@RequiredArgsConstructor
+public class InitialContributionJobConsumerContainer {
+
+  public static final long POLL_TIMEOUT = 9000000L;
```
steveellis: This is 2.5 hours. It seems quite a bit higher than it needs to be. My understanding is:
* This value [needs to be greater](https://stackoverflow.com/questions/66108816/spring-kafka-2-6-5-infinite-retry-policy-via-stateful-retry-and-seektocurrenterr) than what we're setting as the `kafka.backoff.interval` (used as `retryConfig.getInterval`) but 2.5 hours is a big number.
* [The default of this value is 5 minutes](https://docs.spring.io/spring-kafka/api/org/springframework/kafka/listener/ConsumerProperties.html#setPollTimeout(long)).

@Arghya22 it seems like you think that 5 minutes isn't enough (if it is 5 minutes, the default). Was something else overriding the 5 minute default perhaps?

Is there a downside to keeping it at 2.5 hours? I guess that it could hide a broken Kafka retry process, keeping the container alive when we in fact don't want it to be stuck retrying, but I'm not sure.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -333,7 +415,11 @@ private void deContributeInstance(UUID centralServerId, Instance instance, Stati
...
       recordContributionService.deContributeInstance(centralServerId, instance);
       stats.addRecordsDeContributed(1);
-    } catch (Exception e) {
+    }
+    catch (ServiceSuspendedException ex) {
+      throw new ServiceSuspendedException(ex.getMessage());
+    }
+    catch (Exception e) {
       instanceExceptionListener.logWriteError(e, instance.getId());
     } finally {
```
steveellis, line 427: You're not incrementing `recordsProcessed` here. Is that because you feel `contributeItemsChunk` will handle the incrementation?

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -305,24 +364,47 @@ private void contributeItemsChunk(UUID centralServerId, String bibId, List<Item>
       stats.addRecordsTotal(itemsCount);
       recordContributionService.contributeItems(centralServerId, bibId, items);
       stats.addRecordsContributed(itemsCount);
-    } catch (Exception e) {
+      ContributionJobRunner.recordsProcessed.put(getContributionJobContext().getTenantId(), recordsProcessed.get(getContributionJobContext().getTenantId()) == null ? 1
```
steveellis: What is stopping ongoing contributions from incrementing the `recordsProcessed` counter? Because of course if that happens the count will be wrong.

steveellis: @Arghya22 is it that the `ThreadLocal` of the `ContributionJobContextManager` handles this? cc @gurleenkaurbp

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Arghya22: this is part of the **deContrbution** which in this PR or in this story not handled. This PR is all about initial contribution.

src/main/java/org/folio/innreach/batch/contribution/InitialContributionJobConsumerContainer.java
Arghya22: default time 5 min is enough and this big number will create no effect as now kafka itself is handling the retry with a interval of 90 sec . As a safety measure I added it , and surely it wont create any issue as its a max value.

src/main/java/org/folio/innreach/config/props/InnReachRetryPolicy.java
Arghya22: It won't cause issue and for 404 error spring retry will trigger. 

why ? : 
1. When 404 error occurs first **InnReachFeignErrorDecoder** gets triggered and it will be part of **InnReachException** which will be thrown.
2. Now after that **InnReachRetryPolicy** else-if statement will be executed and if you see there is no Classifiable which is handling **InnReachException** , so s**pring retry Template** will get triggered properly for **404 error**. 
3. 
image

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Arghya22: should we do **onGoing** contribution ( updating /creating marc records ) when initialContribution is running ? Why someone wants to change it when they ( tenant ) already triggered initial Contribution !

steveellis: Unfortunately it looks like stopping ongoing contributions isn't an option. We can't ask them to stop things like performing checkouts and such. So we need to fix this ASAP and get a new build to Carole.

Arghya22: checks are added to avoid incrementation of count for ongoing contribution. But having said this , ongoing contrbution retry fix is not part of this jira/PR , so testing should be done only for **Initial contribution** and ongoing contribution ( eg. update / add MARC instance,holding or item should be avoided while testing this PR ) . All fix of ongoing contrbution will be part of **[MODINREACH-351](https://issues.folio.org/browse/MODINREACH-351)**

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -62,86 +61,141 @@ public class ContributionJobRunner {
...
+      contributeInstanceItems(centralServerId, instance, stats);
+    } else if (isContributed(centralServerId, instance)) {
+      log.info("Initial: deContributeInstance");
+      deContributeInstance(centralServerId, instance, stats);
+    }
+    else {
+      // to test if non-eligible increasing count to verify the stopping condition
+      log.info("Initial: non-eligible instance");
+      ContributionJobRunner.recordsProcessed.put(context.getTenantId(), recordsProcessed.get(context.getTenantId()) == null ? 1
+        : recordsProcessed.get(context.getTenantId())+1);
```
steveellis, line 146: This should be using the new `addRecordProcessed` method shouldn't it?

README.md
```diff
@@ -41,6 +41,8 @@ The module provides an access to INN-Reach.
 | KAFKA_SSL_TRUSTSTORE_LOCATION |             -             | The location of the Kafka trust store file.                                                                                                                |
 | KAFKA_SSL_TRUSTSTORE_PASSWORD |             -             | The password for the Kafka trust store file. If a password is not set, trust store file configured will still be used, but integrity checking is disabled. |
 | LOG_HTTP                      |           false           | Enable logging of all requests and responses.                                                                                                              |
+| DEFAULT_INTERVAL              |           95000           | Default interval in ms ( preffred not to increase more than this value )                                                                                   |
+ | MAX_FAILURE                   |     360                   | Default max attempts                                                                                                                                       |
```
steveellis, line 45: See my comment elsewhere about these. These will need to be updated too.

src/main/resources/application.yml
```diff
@@ -168,3 +168,8 @@ kafka:
       topic-pattern: (${environment}\.)(.*\.)inventory\.holdings-record
       concurrency: ${KAFKA_EVENTS_CONCURRENCY:2}
       data-type: org.folio.innreach.dto.Holding
+  backoff:
+    interval: ${DEFAULT_INTERVAL:95000}
+    max_failure: ${MAX_FAILURE:360}
```
steveellis, line 173: Noticing that there is a section in the application.yml called `batch.jobs.contribution` which has items like `retry-attempts` and `retry-interval-ms`. These properties are used in the `ContributionJobProperties` as `retryAttempts` and `retryIntervalMs`. Are these being used?

If not, it seems like we need to do one of two things:
1. Remove the `batch.jobs.contribution` section and related code completely if it isn't being used.
2. Remove the `kafka.backoff` section of the yaml and instead use the `batch.jobs.contribution` properties instead. Seems like the word "batch" is used elsewhere as a namespace for "initial contribution" job stuff. So this might be better.

src/main/resources/application.yml
```diff
@@ -168,3 +168,8 @@ kafka:
...
+  custom-offset: ${DEFAULT_OFFSET:latest}
+  custom-concurrency: ${DEFAULT_CONCURRENCY:2}
```
steveellis, line 175: If we to keep the `batch.jobs.contribution` it probably would make more sense to put this concurrency level there and just call it `batch.jobs.contribution.concurrency`.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Arghya22: addRecordProcessed () is used to restrict onGoing contribution from increasing count and used in methods which is used by both initial and onGoing contribution . In this case not required as this method is only used by initial contribution only.

src/main/resources/application.yml
Arghya22: batch.jobs.contribution propery has no relation with custom kafka related properties . That is used in retryTemplate configurations class. As per me custom kafka property should stay in one place .  which helps to pull the data as well in code

src/main/resources/application.yml
Arghya22: but this is kafka related properties and has no relation with batch.jobs.contribution properties.

steveellis (approved):
Thanks @Arghya22 for all your work on this!

gurleenkaurbp (approved):

Vignesh-kalyanasundaram (approved):

PR #353, opened 2023-07-27
MODINNREACH-380 Implement Outbox pattern for storing and processing of Kafka events

Vignesh-kalyanasundaram:
## Purpose
https://issues.folio.org/browse/MODINREACH-380 The purpose of this jira is to implement the outbox pattern for storing and processing Kafka events for Initial contribution

## Approach
we created a new table JOB_EXECUTION_STATUS and persist the Kafka events with status as READY and instance_contributed as false. The events will be processed periodically using Spring scheduler. we will fetch the items which are having the status Ready and Retry. If retry is 0, then we will consider it as a indefinite retry or else we will check the retry_attempt exceeds max_attempts. If we get any of the following exception ServiceSuspendedException, InnReachConnectionException, FeignException, SocketTimeOutExceptionWrapper , then we will update the status as RETRY. If the instance_contributed  is false, then we will contribute the instance or else we will contribute the items. In order to skip the verification of instance, we keep interval in the query where we can mention the pause between instance and item contribution. The initiate contribution status is updated in the contribution table whenever the user calls contribution/current API. 

#### TODOS and Open Questions

- [ ] Check logging

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/batch/contribution/listener/ContributionExceptionListener.java
```diff
@@ -43,6 +43,18 @@ public void logReaderError(Exception e) {
   public void logWriteError(Exception e, UUID recordId) {
     logError(e, defaultIfNull(recordId, UNKNOWN_ID), "write");
   }
+  public void logError(Exception e, UUID recordId, UUID contributionId) {
```
steveellis, line 47: Need space above method here

src/main/java/org/folio/innreach/repository/JobExecutionStatusRepository.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.UUID;
+
+@Repository
+public interface JobExecutionStatusRepository extends JpaRepository<JobExecutionStatus, UUID> {
+
+  @Query(value = "update job_execution_status s " +
+    "set status = 'IN_PROGRESS' where s.id in (select t.id from job_execution_status t " +
+    "inner join contribution c on c.job_id = t.job_id where c.status = 0 and " +
+    "((t.instance_contributed = true and t.updated_date < current_timestamp - (interval '1 hour') * :itemPause) " +
+    "or t.instance_contributed = false) and "+
```
steveellis: Need space between `"+` characters

src/main/java/org/folio/innreach/domain/entity/JobExecutionStatus.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@Entity
+@Table(name = "job_execution_status")
+@ToString
+@EqualsAndHashCode(of = "id", callSuper = false)
+public class JobExecutionStatus extends Auditable {
```
steveellis, line 26: I wonder should this be called something like `InstanceContributionStatus` or maybe just `ContributionStatus` since there are many instances/contributions being contributed for one job id I think.

src/main/java/org/folio/innreach/ccheduler/InitialContributionJobScheduler.java
```diff
@@ -0,0 +1,62 @@
...
+  @Scheduled(fixedDelayString = "${initial-contribution.scheduler.fixed-delay}",
+    initialDelayString = "${initial-contribution.scheduler.initial-delay}")
+  public void processInitialContributionEvents() {
+    List<String> tenants = loadTenants();
+    log.info("processInitialContributionEvents :: tenantsList {}", tenants);
+    tenants.forEach(tenant ->
+      executionService.runTenantScoped(tenant,
+        () -> {
+          try {
+            jobExecutionStatusRepository.updateAndFetchJobExecutionRecordsByStatus(recordLimit, itemPause)
```
steveellis: Just want to make sure I understand how concurrency is working for us here. Does a `recordLimit` of `2` mean we will have a concurrency level of 2, meaning two threads/cores will be working simultaneously. I believe this, paired with `@Async` on `processInitialContributionEvents` will have this effect but correct me if I'm wrong.

Vignesh-kalyanasundaram: Yes steve, you are correct. We will fetch `recordLimit ` number of records from the database and process every record asynchronously.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
```diff
@@ -0,0 +1,178 @@
...
+        log.warn("processInitialContributionEvents:: Retrying the contribution for instanceId {} due to {} ",
+          job.getInstanceId(), ex.getMessage());
+        updateJobAndContributionStatus(job, RETRY, job.isInstanceContributed());
+      } catch (Exception ex) {
+        log.warn("processInitialContributionEvents:: Exception while processing instanceId {} ", job.getInstanceId());
+        logException(job, ex, contributionRecord.get(job.getJobId()).getId());
+        updateJobAndContributionStatus(job, FAILED, job.isInstanceContributed());
+      }
+    });
+  }
```
steveellis: Think if we have any ability to add additional details to the log messages here, we should take it before deploying -- speaking from experience trying to debug this in production. For example, I'm not seeing the job id being logged. Basically all the details that can be added should be added.

steveellis: It looks like `JobExecutionStatus` is in scope here and it would have all the details that we might want in each of these log messages for `warn` or `info`. It looks like that is annotated with lombok's `ToString` annotation which should be called by log4j here (at least in the `info` line) into something meaningful. Please LMK if that is correct.

If that's true, then perhaps all of this logging doesn't need to be changed. The instance id is logged in each place and that should be enough to tie it back to the particular `JobExcecutionStatus`. See if you agree.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
```diff
@@ -0,0 +1,178 @@
...
+      contributionRecord.put(jobId, contribution);
+      return contribution.getCentralServer().getId();
+    }
+    return null;
+  }
+
+  private void startContribution(UUID centralServerId, Instance instance, JobExecutionStatus job) throws SocketTimeoutException {
+    var instanceId = job.getInstanceId();
+    if (isEligibleForContribution(centralServerId, instance)) {
+      if(job.isInstanceContributed()){
```
steveellis: Need space between `){` chars

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
```diff
@@ -0,0 +1,178 @@
...
+        log.info("startContribution:: Item contribution started for centralServerId: {}, instanceId: {}", centralServerId, instanceId);
+        contributeItem(job, centralServerId, instance);
+      } else {
+        log.info("startContribution:: Instance contribution started for instanceId {} ", instance.getId());
+        recordContributionService.contributeInstanceWithoutRetry(centralServerId, instance);
+        updateJobAndContributionStatus(job, READY, true);
+      }
+    } else if (isContributed(centralServerId, instance)) {
```
steveellis, line 117: Let's discuss. Wouldn't this have the effect of de-contributing anything that had been contributed previously. We're running this to "shore up" an existing initial contribution job. Not sure if this is what we want. I will talk to Brooks about this. But let's make sure to discuss before we roll this out.

steveellis: Ok I verified that this is in fact the correct logic with Brooks. The contribution eligibility can have changed since it was last contributed, and if it is no longer eligible, and has been contributed, it should be de-contributed.

src/main/java/org/folio/innreach/ccheduler/InitialContributionJobScheduler.java
```diff
@@ -0,0 +1,62 @@
...
+public class InitialContributionJobScheduler {
+  private final TenantScopedExecutionService executionService;
+  private final JobExecutionStatusRepository jobExecutionStatusRepository;
+  private final InitialContributionEventProcessor eventProcessor;
+  private final TenantInfoRepository tenantRepository;
+  @Value(value = "${initial-contribution.fetch-limit}")
+  private int recordLimit;
+  @Value(value = "${initial-contribution.item-pause}")
+  private int itemPause;
+  private final Cache<String, List<String>> tenantDetailsCache;
```
steveellis: I think we need to add an `AtomicInteger` property called `currentRequestCount` or similar here which can be used to track the count of contribution requests that are in the state of being in progress. This can be incremented on start and de-incremented on success or failure. This integer can be passed as an argument to the `@Async` method when it is called each time.

This will allow us to set the scheduler interval to be very low, like say 100 milliseconds. Each time it fires we first check the the `currentRequestCount` and compare it to config value we already have for the fetch limit and if there are no "slots" available (`fetchLimit == currentRequestCount`), return and wait for the next interval to pass. If a slot is available, increment the counter and start the request.

If we don't have something like this here is what I think will happen:
* If the interval is too brief the OS thread scheduler will spawn a new thread each time it fires, which will eventually cause OOM unless we somehow get the interval "just right".
* If the interval is too long, we may not get OOM (at least not right away), but performance will suffer since it won't always be doing work processing requests.

Even in the case when the interval is set to a "long time", we could still get ourselves into the first situation if inn reach happens to be performing really slow for a period of time. Lots of requests will get queued up each time it fires, and if inn-reach is really really slow, the same OOM situation will eventually occur.

Anyway I think this is an easy problem to fix. See what you think.

cc @kalibek-epam

kalibek-epam: @steveellis the proper interval and amount of records per single invocation is the key part here. They are configurable parameters and their goal is to allow flexibility to find values to avoid the OOMs. On the other hand, what you propose is effectively introducing state inside stateless service. The goal of the outbox approach was to avoid it. Whatever state we may need like `currentRequestCount` should be persisted in the database.

steveellis: @kalibek-epam We could use the database to get the `currentRequestCount` instead of the `AtomicInteger` bound to the service. I know that we have the in progress state represented there in the db.

But I think getting this from the db makes the app harder to reason about. Consider the following.

If we have 2 containers running our app, container A and B, and each container has a CPU capacity of 4 vCPUs. Fetch the `currentRequestCount` from the db.  At any given time the `currentRequestCount` will be <= 8, the total count across all app instances. Now let's say container A happens to have one of its requests taking a really long time. Will the `currentRequestCount` if obtained from the database represent this state?

To put it another way: how does the container instance translate this count to its local state, which is what we need to know. Will the `fetchLimit` be equal to the sum of CPU capacity of all containers? It could be set to that. And if so container A would see that it doesn't have capacity to queue the next request, but so would container B. Also each time we add a container to the cluster of containers, we need to update this configurable `fetchLimit`.

How could this go wrong? Lets imagine one or two long running requests lasts 20 seconds, and our interval is firing every 100 milliseconds. If the container doesn't have real awareness of its own requests, this could mean that the scheduler would queue 20 * 10 requests prematurely in container A, which puts us back in the same situation we're trying to avoid.

I think all of these problems go away if we just track the container's own request count inside of the container. The most useful count here is the count as it relates to each container's cpu capacity.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
```diff
@@ -0,0 +1,178 @@
...
+      recordContributionService.deContributeInstance(centralServerId, instance);
+      updateJobAndContributionStatus(job, DE_CONTRIBUTED, job.isInstanceContributed());
+    } else {
+      // Update the status of non-eligible instance Id
+      log.info("startContribution:: non-eligible instance id: {}", instanceId);
+      updateJobAndContributionStatus(job, FAILED, job.isInstanceContributed());
+    }
+  }
+
+  private void contributeItem(JobExecutionStatus job, UUID centralServerId, Instance instance) {
```
steveellis: This should probably be called `contributeItems` since there usually more than one.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
steveellis: I'm trying to imagine what it will be like keeping tabs on this job while it is running. I know we're wanting to remove the statistics because the design in the `ContributionJobRunner` was stored in the app state. We don't want to go back to that.

And yet we still have a need to keep statistics of some kind, even if it is just log messages telling us how many messages have been processed. **How else will we know that things are working?** I think there is also an UI that relies on the statistics table which the end user uses to get this status.

How can we solve for this?

One idea I know we have talked about is to use the outbox table as the source of truth. We could query this while the job runs, running a `count` on outbox the table, for various statuses, perhaps when updating in `updateJobAndContributionStatus`, and then update the statistics table in the DB with the current counts of each state. Because as it stands now I think the status will always show 0 for the job unless I'm missing something. We don't have to do it for every status update necessarily (it doesn't have to be perfect), just roughly correct.

We already have `ContributionJobStatsListener.updateStats` which writes the stats to storage. We could just populate this with stats that are the result of our `count` sql against the job table. We could also log this information to an `info` statement so we can keep tabs on it in CW.

src/main/java/org/folio/innreach/ccheduler/InitialContributionJobScheduler.java
steveellis: I think we agreed today to try it with the db-oriented approach. Let's see how it goes.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
steveellis: I didn't notice that the stats were being upgraded elsewhere.

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
Vignesh-kalyanasundaram: Updated to contributeItems

src/main/java/org/folio/innreach/batch/contribution/service/InitialContributionEventProcessor.java
Vignesh-kalyanasundaram: Space added

src/main/java/org/folio/innreach/repository/JobExecutionStatusRepository.java
Vignesh-kalyanasundaram: space added

src/main/java/org/folio/innreach/batch/contribution/listener/ContributionExceptionListener.java
Vignesh-kalyanasundaram: Space added

gurleenkaurbp (approved):

PR #359, opened 2023-08-22
MODINREACH-391

Vignesh-kalyanasundaram:
## Purpose
The purpose of the PR is to fix https://issues.folio.org/browse/MODINREACH-391.

## Approach
During ongoing contribution, in order to reuse the existing contribution , We are fetching Ongoing contributions which is having status as 0(In_Progress) and ongoing value as true. This is failing when there are more than 1 row which is having the above mentioned values in DB. The failure is because we are trying to wrap **list of contribution** into **contribution**. In order to fix this issue, we are removing the existing contribution check so that for every contribution there will be a new row which gets created in DB.

#### TODOS and Open Questions

- [ ] Check logging

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/innreach/domain/service/impl/ContributionServiceImpl.java
+++ b/src/main/java/org/folio/innreach/domain/service/impl/ContributionServiceImpl.java
@@ -10,7 +10,6 @@ import static org.folio.innreach.dto.MappingValidationStatusDTO.VALID;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 import java.util.UUID;
 
 import lombok.RequiredArgsConstructor;
@@ -185,11 +184,6 @@ public class ContributionServiceImpl implements ContributionService {
   public ContributionDTO createOngoingContribution(UUID centralServerId) {
     log.info("createOngoingContribution:: parameters centralServerId: {}", centralServerId);
     var contribution = createEmptyContribution(centralServerId);
-    Optional<Contribution> oldContribution = repository.fetchOngoingByCentralServerId(centralServerId);
-    if(oldContribution.isPresent()) {
-      log.info("createOngoingContribution :: old contribution");
-      return mapper.toDTO(oldContribution.get());
-    }
     contribution.setOngoing(true);
 
     return mapper.toDTO(repository.save(contribution));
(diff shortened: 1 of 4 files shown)


nirmalr-shah (approved):

src/main/java/org/folio/innreach/domain/service/impl/ContributionServiceImpl.java
```diff
@@ -185,11 +184,6 @@ private JobResponse getJobResponse(UUID id) {
   public ContributionDTO createOngoingContribution(UUID centralServerId) {
     log.info("createOngoingContribution:: parameters centralServerId: {}", centralServerId);
     var contribution = createEmptyContribution(centralServerId);
```
steveellis, line 186: Maybe instead of creating the empt contribution it should create a row that has a count of 1 for the message that we are processing here just to make the stats "correct". Also probably want to rename this method `createEmptyContribution` to something like `createContribution`.

Vignesh-kalyanasundaram: @steveellis In the current code implementation, `createEmptyContribution ` method initialize the stats value to 0 and gets saved. there are methods which increment the count with the given value. Please refer below.

```
public void addRecordsProcessed(int itemsCount) {
      recordsProcessed += itemsCount;
    }
```

Let's consider below 2 different scenarios where the stats will be different.
1) Instance Update - In this case , only instance contribution can happen. 
records_total - 1, records_processed - 1 , records_contributed - 1, records_decontributed - 0
2) Item delete- In this case , item decontribution and instance contribution can happen. 
records_total - 2, records_processed - 2 , records_contributed - 1, records_decontributed - 1

So, If we initialize to 1, then the stats will be showing incorrect. 
If you agree with the above, I think it is better to keep the name of the method as createEmptyContribution as we always initialize the stats related column value as 0.

pavankumar181 (approved):

gurleenkaurbp (approved):

src/main/java/org/folio/innreach/domain/service/impl/ContributionServiceImpl.java
steveellis: Ok yeah I see your point. The count isn't going to be right in all cases so may as well just leave it as is.

steveellis (approved):

PR #390, opened 2024-03-18
MODINREACH-424: Patron validation endpoint should strip commas and leading and trailing spaces from the patron name field

steveellis:
https://folio-org.atlassian.net/browse/MODINREACH-424

The ticket was reopened and I added a new requirement to it which is to strip leading and trailing spaces and changed the title to reflect that. This PR covers only removing the leading and trailing spaces.

### DIFF
--- a/src/main/java/org/folio/innreach/domain/service/impl/PatronInfoServiceImpl.java
+++ b/src/main/java/org/folio/innreach/domain/service/impl/PatronInfoServiceImpl.java
@@ -12,7 +12,11 @@ import static org.folio.innreach.external.dto.InnReachResponse.Error.ofMessage;
 
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Date;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/innreach/domain/service/impl/PatronInfoServiceImpl.java
```diff
@@ -12,9 +12,7 @@
 
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-import java.util.Date;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
gurleenkaurbp: Please remove the wildcard imports.

gurleenkaurbp (approved):

gurleenkaurbp:
Can we have some description details in the PR for us to fall back to if required in the future.

steveellis:
> Can we have some description details in the PR for us to fall back to if required in the future.

Added a description and a title. Changed the title to reflect the new requirement.

PR #393, opened 2024-05-30
MODINREACH-431 Implementing Outbox pattern for processing ongoing contribution item events

Vignesh-kalyanasundaram:
## Purpose
The purpose of the PR is to implement [MODINREACH-431](https://folio-org.atlassian.net/browse/MODINREACH-431)

## Approach
We created a new table and decided to persist all the ongoing events. As part of this PR, we are persisting item related events into the table. Since we need to process the events for each central server, we are adding the entry against all the central server id in the database. For example, if we have 3 central servers, then for a single event we are adding 3 entries. By this way if any error happens, we can easily identify the central server which is not able to process the event and we can reprocess the message if it throws the list of retryable exception. Processing logic of event is not changed. Removed the code of itemDelete and itemUpdate as it is already implemented in this PR. Since itemcreate method is reused by some method, it will be removed in future when there is no dependency. Retrieval of records from the database is similar to Initial contribution and we decided to reuse the same params of Initial contribution so those properties in application.yml file are renamed.

#### TODOS and Open Questions

- [ ] Check logging

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -248,45 +253,81 @@ public void runItemContribution(UUID centralServerId, Instance instance, Item it
...
+      log.info("Ongoing: validating item {} for contribution to central server {} with instance id: {}", item.getId(), centralServerId, instance.getId());
+      boolean eligibleItem = isEligibleForContribution(centralServerId, item);
+      boolean contributedItem = isContributed(centralServerId, instance, item);
+      log.info("Ongoing: eligibleItem: {}, contributedItem: {}", eligibleItem, contributedItem);
+      if (!eligibleItem && !contributedItem) {
+        log.info("Ongoing: skipping ineligible and non-contributed centralServer id: {}, item id: {}, instance id : {}", centralServerId, item.getId(), instance.getId());
+        ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, SKIPPING_INELIGIBLE_MSG, FAILED);
+        return;
+      }
+      if (isEligibleForContribution(centralServerId, instance)) {
```
gurleenkaurbp, line 267: I think you can use the boolean flag eligibleItem instead of calling the method isEligibleForContribution

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -248,45 +253,81 @@ public void runItemContribution(UUID centralServerId, Instance instance, Item it
...
+        log.info("Ongoing: Re-contributing instance to update bib status, centralServer id: {}, instance id : {}, item id: {}", centralServerId, instance.getId(), item.getId());
+        recordContributionService.contributeInstance(centralServerId, instance);
+        if (eligibleItem) {
```
gurleenkaurbp, line 270: Not sure if this if condition is again needed ? I think the outside if is probably doing the same thing ?

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+    } catch (ServiceSuspendedException | InnReachConnectionException |
+             SocketTimeOutExceptionWrapper | InnReachGatewayException ex) {
+      ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, RETRY);
+    } catch (Exception ex) {
+      ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, ex.getMessage(), FAILED);
+    }
+  }
+
+  private void processItem(OngoingContributionStatus ongoingContributionStatus) {
+    Item oldEntity = jsonHelper.fromJson(ongoingContributionStatus.getOldEntity(), Item.class);
```
gurleenkaurbp, line 66: A logger here might help.

src/main/java/org/folio/innreach/domain/listener/KafkaInventoryEventListener.java
```diff
@@ -146,4 +142,10 @@ else if (event.getType().equals(DELETED)) {
       }
     }
   }
+
+  private List<UUID> getCentralServerIds() {
+    Page<UUID> ids = centralServerRepository.getIds(new OffsetRequest(0, 2000));
```
gurleenkaurbp, line 154: Do we want to make this value 2000 configurable ? Or this is an optimum limit to fetch the records ?

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
```diff
@@ -24,4 +32,23 @@ public <T> void process(T event, Consumer<T> eventProcessor, String tenant) {
...
     }
   }
+
+  @Override
+  public <T> void process(List<DomainEvent<T>> events, BiConsumer<List<DomainEvent<T>>, String> eventProcessor) {
+    try {
+      Map<String, List<DomainEvent<T>>> tenantMap = events.stream().collect(Collectors.groupingBy(DomainEvent::getTenant));
+      tenantMap.forEach((tenant, eventList) -> {
+        log.info("tenant {}", tenant);
+        log.info("eventList for tenant {}", eventList);
```
gurleenkaurbp: I believe these logs will be debug

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Vignesh-kalyanasundaram: boolean variable eligibleItem is evaluating item but here we are checking instance is eligible

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Vignesh-kalyanasundaram: The outside condition is evaluating instance is eligible and this condition is checking item is eligible

src/main/java/org/folio/innreach/domain/listener/KafkaInventoryEventListener.java
Vignesh-kalyanasundaram: I think this is fine. our need is to fetch all records of centralserverIds so that we can add a event against each centralServerid in database

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: Fixed

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
Vignesh-kalyanasundaram: changed to debug

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+            case ITEM -> processItem(ongoingContributionStatus);
+            case HOLDINGS -> processHoldings(ongoingContributionStatus);
+            case INSTANCE -> processInstance(ongoingContributionStatus);
+            default ->
+              ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, UNKNOWN_EVENT_NAME_MESSAGE, FAILED);
+          }
+        });
+    } catch (ServiceSuspendedException | InnReachConnectionException |
+             SocketTimeOutExceptionWrapper | InnReachGatewayException ex) {
+      ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, RETRY);
```
steveellis, line 58: Should we log these exceptions? I think yes. Even if it is `ServiceSuspendendException` we would want to know when and how many times it occurred. Same for other types.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+    } catch (Exception ex) {
+      ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, ex.getMessage(), FAILED);
```
steveellis, line 61: Definitely want to log this one, with a stack trace also so we can know exactly where it is failing.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+        ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, UNKNOWN_TYPE_MESSAGE, FAILED);
+    }
+  }
+
+  private void processHoldings(OngoingContributionStatus ongoingContributionStatus) {
+    log.info("Not yet implemented {}", ongoingContributionStatus);
+  }
+
+  private void processInstance(OngoingContributionStatus ongoingContributionStatus) {
+    log.info("Not yet implemented {}", ongoingContributionStatus);
```
steveellis: I would add some prefix to both of these log messages to distinguish which is being called.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private void checkRetryLimit(OngoingContributionStatus job) {
+    if (maxRetryAttempts != 0 && job.getRetryAttempts() > maxRetryAttempts) {
+      log.warn("checkRetryLimit:: Retry limit exhausted for jobId {} ", job.getId());
+      throw new RetryException(RETRY_LIMIT_MESSAGE);
```
steveellis, line 93: Are we catching this or is it flowing down to the more general exception `Exception` handler?

steveellis: Should you perhaps add the attempts and maxRetryAttempts to the log message. This will help with understanding why it occurred.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
steveellis: I would add the job id to the log messages and any other helpful info.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionStatusServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+  public void updateOngoingContribution(OngoingContributionStatus ongoingContributionStatus,
+                                        ContributionStatus status) {
+    updateStatusAndSaveOngoingJob(ongoingContributionStatus, status);
+  }
+
+  private void updateStatusAndSaveOngoingJob(OngoingContributionStatus ongoingContributionStatus,
+                                             ContributionStatus status) {
+    ongoingContributionStatus.setStatus(status);
+    ongoingContributionStatus.setRetryAttempts(status.equals(RETRY) ?
+      ongoingContributionStatus.getRetryAttempts() + 1 : ongoingContributionStatus.getRetryAttempts());
```
steveellis, line 35: Minor but do you need to get the retry attempts here? Won't it already have the right value. Maybe ternary should be replaced by a `if`.

src/main/java/org/folio/innreach/config/SpringAsyncConfig.java
```diff
@@ -47,13 +47,24 @@ public ThreadPoolTaskScheduler prepareScheduler() {
...
+    executor.initialize();
+    return executor;
+  }
+
+  @Bean("ongoingSchedulerTaskExecutor")
+  public Executor ongoingSchedulerTaskExecutor() {
+    ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(schedulerTaskPoolSize);
+    executor.setMaxPoolSize(schedulerTaskPoolSize + 10);
+    executor.setQueueCapacity(50);
```
steveellis, line 66: Should these be configurable? Are they configurable for initial contributions? Not saying they should be but maybe if they are for initial.

src/main/java/org/folio/innreach/domain/listener/KafkaInventoryEventListener.java
steveellis: @Vignesh-kalyanasundaram  can you say some more about why it shouldn't be configurable? Not sure I understand your comment.

src/main/java/org/folio/innreach/domain/service/impl/ContributionActionServiceImpl.java
```diff
@@ -95,45 +101,58 @@ public void handleItemCreation(Item newItem) {
...
+  public void handleItemCreation(Item newItem, OngoingContributionStatus ongoingContributionStatus) {
+    log.info("Handling item creation {}", newItem.getId());
+
+    var instance = fetchInstanceWithItems(newItem);
+    if (!isMARCRecord(instance)) {
+      ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, MARC_ERROR_MSG, FAILED);
+      return;
+    }
+    var centralServerId = ongoingContributionStatus.getCentralServerId();
+    if(checkCentralServerValid(centralServerId)) {
```
steveellis: space after `if`

src/main/java/org/folio/innreach/domain/service/impl/ContributionActionServiceImpl.java
```diff
@@ -95,45 +101,58 @@ public void handleItemCreation(Item newItem) {
...
       return;
     }
-    try {
-      for (var csId : getCentralServerIds()) {
-        contributionJobRunner.runItemDeContribution(csId, instance, deletedItem);
-      }
-    } catch (ServiceSuspendedException | FeignException | InnReachConnectionException e) {
-      log.info("exception thrown from handleItemDelete", e);
-      throw e;
-    }
```
steveellis, line 136: I think moving these to the highest level where they could occur is the right choice.

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
```diff
@@ -24,4 +32,23 @@ public <T> void process(T event, Consumer<T> eventProcessor, String tenant) {
...
+    try {
+      Map<String, List<DomainEvent<T>>> tenantMap = events.stream().collect(Collectors.groupingBy(DomainEvent::getTenant));
+      tenantMap.forEach((tenant, eventList) -> {
+        log.debug("tenant {}", tenant);
+        log.debug("eventList for tenant {}", eventList);
+        if (innReachTenants.contains(tenant)) {
+          executionService.runTenantScoped(tenant,
+            () -> eventProcessor.accept(eventList, tenant));
+        } else {
+          log.warn("Ignoring event of unknown tenant {}", tenant);
```
steveellis: Should this be a `warn`. Isn't it the expected behavior that unknown tenants will arrive here? Seems like an `info` maybe.

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
```diff
@@ -24,4 +32,23 @@ public <T> void process(T event, Consumer<T> eventProcessor, String tenant) {
...
+        }
+      });
+    } catch (Exception ex) {
+      log.warn("Unable to save kafka event into outbox table ", ex);
```
steveellis: Should this be an `error` log and maybe also add details or event itself to the message.

src/main/resources/application.yml
```diff
@@ -180,18 +180,18 @@ kafka:
       data-type: org.folio.innreach.domain.dto.folio.inventorystorage.InstanceIterationEvent
   backoff:
     interval: ${DEFAULT_INTERVAL:95000}
-    max_failure: ${MAX_FAILURE:360}
+    max_failure: ${MAX_FAILURE:1}
```
steveellis: Do we need to have an application parameter for setting a limit on the retries during suspension? Is that what is here? We may want to retry indefinitely now that we have the outbox. Seems like there is`retry_attempts` below. 

Or is it that we're planning to get rid of `backoff` entirely with future changes to remove the old code?

src/main/resources/application.yml
```diff
@@ -180,18 +180,18 @@ kafka:
...
   scheduler:
-    fixed-delay: ${INITIAL_CONTRIBUTION_SCHEDULER_DELAY:10000}
+    fixed-delay: ${CONTRIBUTION_SCHEDULER_DELAY:10000}
     initial-delay: 300000
-  retry-attempts: ${INITIAL_CONTRIBUTION_RETRIES:3600}
-  fetch-limit: ${INITIAL_CONTRIBUTION_FETCH_LIMIT:50}
-  item-pause: ${INITIAL_CONTRIBUTION_ITEM_PAUSE:0.5}
+  retry-attempts: ${CONTRIBUTION_RETRIES:3600}
+  fetch-limit: ${CONTRIBUTION_FETCH_LIMIT:50}
+  item-pause: ${CONTRIBUTION_ITEM_PAUSE:0.5}
```
steveellis, line 195: See my comment [above](https://github.com/folio-org/mod-inn-reach/pull/393/files#r1643156094).

src/main/java/org/folio/innreach/domain/service/impl/ContributionActionServiceImpl.java
Vignesh-kalyanasundaram: Actually, we are removing this logic because whenever we are persisting ongoing event, we are persisting the event against every central serverId. So this check is not necessary in this place. hence, the code is removed.

src/main/resources/application.yml
Vignesh-kalyanasundaram: retry-attempts is the parameter which is responsible for number of retries. we used this InitialContributionEventProcessor and OngoingContributionEventProcessor class

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionStatusServiceImpl.java
Vignesh-kalyanasundaram: The purpose of getting the retry attempt is to increase the existing retry count by 1 whenever we get the status RETRY, we need to make sure the retry count is incremented. Retry attempts value indicates no of attempts in which we are trying to process a particular record.

src/main/java/org/folio/innreach/config/SpringAsyncConfig.java
Vignesh-kalyanasundaram: we only had CorePoolSize as  configurable. Do you want queueCapacity, maxPoolSize as a configurable one ?

src/main/java/org/folio/innreach/domain/listener/KafkaInventoryEventListener.java
Vignesh-kalyanasundaram: I don't mean that it shouldn't be configurable. I mean to say that we need to fetch all the centralServerIds before saving the event into DB. As the existing code implemented in this manner, ([please see here](https://github.com/folio-org/mod-inn-reach/blob/master/src/main/java/org/folio/innreach/domain/service/impl/ContributionActionServiceImpl.java#L255)), I followed the same

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: We are throwing RETRY Exception when the retry attempts are exceeded and it will be caught by the general exception catch block. In catch block, status of the job is updated to FAILED.

Vignesh-kalyanasundaram: Logger message is modified

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: logger is added

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: logger is added

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: Done

src/main/java/org/folio/innreach/domain/service/impl/ContributionActionServiceImpl.java
Vignesh-kalyanasundaram: Done

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
Vignesh-kalyanasundaram: Yes, changed to info

src/main/java/org/folio/innreach/domain/service/impl/KafkaEventProcessorServiceImpl.java
Vignesh-kalyanasundaram: Yes, changed to error

src/main/resources/application.yml
Vignesh-kalyanasundaram: This is modified wrongly. Now putting the old value itself. As of now, even though  this param is being used in kafka retry config, this wont make any impact as the list of retryable exception will not occur at the time of persisting kafka events into DB

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -248,45 +253,81 @@ public void runItemContribution(UUID centralServerId, Instance instance, Item it
...
+          log.info("Ongoing: contributing centralServer id:{}, instance id : {}, item id: {}", centralServerId, instance.getId(), item.getId());
+          recordContributionService.contributeItems(centralServerId, instance.getHrid(), List.of(item));
+        } else if (contributedItem) {
+          log.info("Ongoing: de-contributing centralServer id: {}, instance id : {}, item id: {}", centralServerId, instance.getId(), item.getId());
+          recordContributionService.deContributeItem(centralServerId, item);
+        }
+        ongoingContributionStatusService.updateOngoingContribution(ongoingContributionStatus, PROCESSED);
+      } else if (contributedItem) {
+        log.info(" Ongoing : " + DE_CONTRIBUTE_INSTANCE_MSG+", centralServer id: {}, instance id : {}, item id: {}", centralServerId, instance.getId(), item.getId());
```
steveellis: Should there be a `+` after `_MSG`?

steveellis: Probably don't want a space before `Ongoing : `

src/main/java/org/folio/innreach/config/SpringAsyncConfig.java
steveellis: I think what we had for initial is fine. If it was only `CorePoolSize` then that's probably good enough for now.

steveellis (approved):
Looks good Vignesh. Thanks for your work on this!

gurleenkaurbp (approved):

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Vignesh-kalyanasundaram: modified the info message

PR #394, opened 2024-05-30
MODINREACH-435

Vignesh-kalyanasundaram:
## Purpose
The purpose of the PR is to fix [MODINREACH-435](https://folio-org.atlassian.net/browse/MODINREACH-435)

## Approach
Since there are no tenant specific topics from Poppy, we are listening to all the events but filtering the events before processing. As we have the list of tenants where the innreach module is enabled (INNREACH_TENANTS env variable) , we used it to filter the topics.
 Also just to consume all the un processed events, set up the auto-offset-reset to earliest. 
Added a try catch block in kafka error handler, so that if we encounter any exception in the error handler method, it can be handled gracefully.

#### TODOS and Open Questions

- [ ] Check logging

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/config/KafkaListenerConfiguration.java
```diff
@@ -121,9 +121,13 @@ public DefaultErrorHandler errorHandler() {
...
       // logic to execute when all the retry attempts are exhausted
-      executionService.runTenantScoped(getContributionJobContext().getTenantId(),
+      try {
+        executionService.runTenantScoped(getContributionJobContext().getTenantId(),
           () -> contributionJobRunner.completeContribution(getContributionJobContext()));
-      endContributionJobContext();
+        endContributionJobContext();
+      } catch (Exception ex) {
+        log.warn("Exception while processing error handler {} ", ex.getMessage());
+      }
```
steveellis, line 130: I see we want a try catch here just so that we don't trigger retry at all for ongoing. Do I have this right?

src/main/resources/application.yml
```diff
@@ -35,6 +35,7 @@ spring:
     consumer:
       max-poll-records: 20
       group-id: ${ENV:folio}-mod-innreach-events-group
+      auto-offset-reset: earliest
```
steveellis, line 38: Awesome!

src/main/java/org/folio/innreach/config/KafkaListenerConfiguration.java
Vignesh-kalyanasundaram: No steve, this lambda function will gets invoked only when all the retries are exhausted. We still have the retry for ongoing

steveellis (approved):

gurleenkaurbp (approved):

src/main/java/org/folio/innreach/config/KafkaListenerConfiguration.java
steveellis: Ok thanks for clarifying

PR #398, opened 2024-06-27
MODINREACH-433 Implementing Outbox pattern for processing ongoing contribution Instance events                  

Vignesh-kalyanasundaram:
## Purpose
The purpose of the PR is to implement [MODINREACH-433](https://folio-org.atlassian.net/browse/MODINREACH-433)

## Approach
Similar to MODINREACH-431, We are persisting all the ongoing instance events in outbox table. Since we need to process the events for each central server, we are adding the entry against all the central server id in the database. Processing logic of event is not changed. Removed the existing code of runInstanceContribution and runInstanceDecontribution as it is no longer used. 

#### TODOS and Open Questions

- [ ] Check logging

## Pre-Merge Checklist:


src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -178,47 +180,51 @@ public void cancelContributionIfRetryExhausted(UUID centralServerId) {
...
-    runOngoing(centralServerId, (ctx, statistics) -> {
-      log.info("Ongoing: starting ongoing instance contribution job {}, centralServerId: {}", ctx, centralServerId);
-
+  public void runOngoingInstanceContribution(UUID centralServerId, Instance instance, OngoingContributionStatus ongoingContributionStatus) {
+    try {
+      log.info("runOngoingInstanceContribution:: validating instance {} for contribution to central server {}", instance.getId(), centralServerId);
+      boolean eligibleInstance = isEligibleForContribution(centralServerId, instance);
+      boolean contributedInstance = isContributed(centralServerId, instance);
+      log.info("runOngoingInstanceContribution:: eligibleInstance: {}, contributedInstance: {}", eligibleInstance, contributedInstance);
+      if (!eligibleInstance && !contributedInstance) {
```
steveellis, line 189: If these two can be null, perhaps we shouldn't log them in line 188. I think seeing them be null in the logs could be misconstrued as an error to be an error when it fact expected. Perhaps only log what is in 188 in an `else` of this `if`.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
```diff
@@ -178,47 +180,51 @@ public void cancelContributionIfRetryExhausted(UUID centralServerId) {
...
+        log.info("runOngoingInstanceContribution:: skipping ineligible and non-contributed instance");
```
steveellis: Can we log the instance id here? Because having a message about skipping without knowing which one was skipped probably isn't going to help people looking at the logs to troubleshoot a production issue.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
steveellis: I'm sorry I somehow didn't notice that these are _booleans_. Maybe I'm writing too much Javascript where null can be checked this way... Sigh.

Anyway I still think the it might be nice to move 188 into an else clause and perhaps add some context to the message. Like what does it mean when these are both false? Instance is not eligible and has not been contributed. Something like that.

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
```diff
@@ -15,8 +15,9 @@
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
-import org.folio.innreach.dto.Item;
 import org.folio.innreach.dto.Holding;
+import org.folio.innreach.dto.Instance;
+import org.folio.innreach.dto.Item;
```
steveellis: These 3 should probably be grouped with the `org.folio.innreach` items above in alpha order.

src/test/java/org/folio/innreach/domain/listener/KafkaInventoryEventListenerApiTest.java
```diff
@@ -106,24 +80,23 @@ void shouldReceiveInventoryHoldingEvent() {
...
-
-    verify(listener).handleInstanceEvents(eventsCaptor.capture());
-
-    var records = eventsCaptor.getValue();
-    assertEquals(1, records.size());
-
-    var record1 = records.get(0);
-    assertEquals(RECORD_ID.toString(), record1.key());
-    assertEquals(event, record1.value());
+      assertEquals(initialSize+6, ongoingContributionRepository.count()));
```
steveellis: Probably want `initialSize + 6` (spaces)

steveellis (approved):

gurleenkaurbp (approved):

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Vignesh-kalyanasundaram: Hi steve, After this if condition, we do have if, else if and else condition based on this Boolean flag. the logger added in those condition will take care of your concern I feel.

src/main/java/org/folio/innreach/batch/contribution/service/ContributionJobRunner.java
Vignesh-kalyanasundaram: I added it

src/test/java/org/folio/innreach/domain/listener/KafkaInventoryEventListenerApiTest.java
Vignesh-kalyanasundaram: changed

src/main/java/org/folio/innreach/batch/contribution/service/OngoingContributionEventProcessor.java
Vignesh-kalyanasundaram: changed

PR #507, opened 2026-05-21
MODINREACH-582: Make improvements to the initial and ongoing contribution

mukhiddin-yusuf:
### Purpose
Performance and correctness improvements for the **initial contribution** and
**ongoing contribution** flows.

The production trigger was a disk-spill error on the PostgreSQL server:

```
ERROR: could not write to file "base/pgsql_tmp/pgsql_tmp29405.0":
No space left on device
```

Root cause: `job_execution_status` has millions of rows and no indexes on the
columns used by the scheduler claim query (`status`, `job_id`,
`instance_contributed`, `updated_date`). Every scheduler tick (per tenant)
ran a full sequential scan + hash join, with the intermediate result set
spilling to disk. The same problem exists for `ongoing_contribution_status`.

Beyond the disk-spill, several scheduler and processor gaps amplify DB load
and risk data correctness: orphaned `IN_PROGRESS` rows when the executor
queue fills, exception handling running outside the tenant context in the
ongoing processor, and unnecessary JSON deserialization for every event.

#### Database — indexes (highest ROI, zero downtime)

- add partial composite index for initial contribution table's columns: `(job_id, instance_contributed, updated_date)` when the status IN `('READY','RETRY')` - prevents a seq-scan + hash join and directly fixes the disk-spill.
- add partial index on contribution table's `job_id` column when `status = 0` - supports the join from the initial-contribution claim query.
- add partial index on initial contribution's `job_id` when `status = 'IN_PROGRESS'` - makes `getInProgressRecordsCount()` an index-only scan.
- add partial index on ongoing contribution's `created_date` when `status IN ('READY','RETRY')` - eliminates the seq-scan + sort on that table.
- add partial index on ongoing contribution covering index `(id)` when `status = 'IN_PROGRESS'` - supports `getInProgressRecordsCount()` and `updateInProgressToReady()`.

#### Query rewrites — `FOR UPDATE SKIP LOCKED`

Both claim queries rewritten from `UPDATE … WHERE id IN (SELECT … LIMIT n)` to
a CTE pattern with `FOR UPDATE SKIP LOCKED`

#### Scheduler — four layers of back-pressure

1. **`minBatchThreshold`** (`contribution.min-batch-threshold`, default `5`):
   only claim a new batch when `available = recordLimit - inProgressCount ≥ threshold`.
   Batch size is `available`, not the full `recordLimit`. Prevents over-claiming
   when InnReach is slow. Applied to both initial and ongoing schedulers.
2. **`ReentrantLock.tryLock()`** on both scheduled methods: if the previous
   tick is still running, the next tick skips immediately. Prevents overlapping
   ticks with zero external dependencies.
3. **`CallerRunsPolicy`** on both task executors: when the pool+queue is full,
   the scheduler thread processes the overflow task itself instead of throwing
   `RejectedExecutionException`. Prevents orphaned `IN_PROGRESS` rows (the
   core data-safety bug of the original `AbortPolicy`).
4. **`FOR UPDATE SKIP LOCKED`** (query layer): pods always claim disjoint rows.

### Changes Checklist
- [ ] **API Changes**: Document any API paths, methods, request or response bodies changed, added, or removed.
- [ ] **Database Schema Changes**: Indicate any database schema changes and their impact. Confirm that migration scripts were created.
- [ ] **Interface Version Changes**: Indicate any changes to interface versions.
- [ ] **Interface Dependencies**: Document added or removed dependencies.
- [ ] **Permissions**: Document any changes to permissions.
- [x] **Logging**: Confirm that logging is appropriately handled.
- [ ] **Unit Testing**: Confirm that changed classes were covered by unit tests.
- [ ] **Integration Testing**: Confirm that changed logic was covered by integration tests.
- [x] **Manual Testing**: Confirm that changes were tested on local or dev environment.
- [x] **NEWS**: Confirm that the NEWS file is updated with relevant information about the changes made in this pull request.

### Related Issues
[MODINREACH-582](https://folio-org.atlassian.net/browse/MODINREACH-582)
[MODINREACH-587](https://folio-org.atlassian.net/browse/MODINREACH-587)


mukhiddin-yusuf:
Investigations and going through possible options for improving the performance and handling of initial and ongoing contribution flows were done using the AI. However, AI's suggestions were not directly applied, some were just not appropriate or makes sense to apply now (they require separate testing and investigation), some were tuned or corrected and each chosen one was considered separately before applying.

Here are two files documenting the process of applying changes using AI:
- [initial_contribution_improve.md](https://github.com/user-attachments/files/28105193/initial_contribution_improve.md)
- [ongoing_contribution_improve.md](https://github.com/user-attachments/files/28105198/ongoing_contribution_improve.md)

pfilippov-epam (approved):

zakletskyi-epam (approved):

